PR #9, opened 2017-08-31
Added species icon to search results

kevinschaper:


paaatrick:
Do we still need an extra copy of the icon images since the gene page CSS is getting reused?

kevinschaper:
Thanks @paaatrick - definitely didn't need the extra sprite images!

PR #11, opened 2017-08-31
AGR-414 Fix layout of wordpress news pages

paaatrick:


LucyHut:
On a second thought , since I already have an issue that addresses the secondary menu display, I think  I should make this part of that issue @AGR-522

PR #24, opened 2017-09-08
Agr 522 - Move static page's secondary navigation from wordpress to react

LucyHut:
I will clean wordpress after the merge ....


LucyHut:
I already cleaned wordpress pages to remove manual step that created the secondary navigation --- Since I will be on vacation 
the last two weeks of this month, i'd like to have this issue done before I leave and any new bug reported as a separate  issue on Jira

PR #32, opened 2017-09-13
Agr 300 panther link

sibyl229:
Hi there,

This PR extracts Panther URL from crossReferences and adds Panther external link to the orthology section.

I added a set of components (AttributeLabel, AttributeValue, AttributeList) for creating attribute list similar to the basic gene info. The objective is to improve readability of the components that needs to create an attribute list (example use case https://github.com/alliance-genome/agr_ui/blob/03a6aefc6ac22ab1f7bd28ed02540b34dd9b184f/src/components/orthology/pantherCrossRef.js#L16-L26). The reason to define custom components at all is to eventually allow consistent styles to be applied throughout the site. AttributeLabel and AttributeValue component accepts arbitrary child component to ensure generalizability of their usage.

Thanks.


sibyl229:
Thanks @kevinschaper !

PR #50, opened 2017-09-25
AGR-593: Add polyfills to support IE11

paaatrick:
Selectively loading a polyfill just for `Promise` wasn't enough to get the gene page going, and the IE error console wasn't particular enlightening on what was missing after that. But loading polyfills wholesale like this works.


paaatrick:
Yes, that sounds like a really good thing to work on post-1.0! I don't want to saddle you with it just because you have a Windows machine, but testing and debugging on a virtualbox VM kiiiiinda sucks 😞 

Anyway I just made a jira issue and stuck it into release 1.3 (AGR-599)

PR #53, opened 2017-09-26
improve error handling and style

nathandunn:


nathandunn:
This just better handles 404s of data (had to make some server changes).  Before it was hanging with the loading screen.  We want to explicitly say that we were not able to retrieve data.  

(success)
/gene/ZFIN:ZDB-GENE-001103-2

vs 
(fail, we have the locations, but it is not in the GFF properly yet)
/gene/MGI:5446626

vs 
(not in BGI to begin with)
/gene/MGI:94928

src/containers/genePage/genomeFeatureViewer.js
```diff
@@ -100,8 +107,8 @@ class GenomeFeatureViewer extends Component {
                                                                  url={this.externalJBrowseUrl}
                                                                  width={this.props.width}
                                                   /> : ''}
-              {this.state.loadState == 'error' ? 'Unable to load' : ''}
             </a>
+            {this.state.loadState == 'error' ? <div className={style.error}>Unable to retrieve data</div> : ''}
```
paaatrick: Maybe consider using bootstrap's `text-danger` class here instead of defining your own to try and keep style consistent?

nathandunn:
That makes a lot of sense.   @scottcain is going to mangle some additional data today, so I’ll have to untangle that.  I’ll put both fixes in this PR.

PR #61, opened 2017-09-28
Fix red error flash when disease page loads

paaatrick:
"lenth"... i am not a smart man 😢


bryon-czoch:
Join the club.  "sting" instead of "string"...

PR #66, opened 2017-09-29
Integrate ribbon 1.x

sibyl229:


sibyl229:
Hi Nathan, thanks for reviewing.  I just noticed I introduced a bug with the highlighting of ribbon label. Working on fixing it right now. Will make a separate PR.

Sibyl

PR #71, opened 2017-09-29
fixed URL to use secure server

nathandunn:
just fixed it to use the https server instead.


nathandunn:
FYI @cmungall  / @kltm   this is now switched over to https.

nathandunn:
FYI @christabone  The Genome Features should be back up now.

PR #80, opened 2017-10-03
AGR-625 Sort disease cross references

paaatrick:


paaatrick:
@sibyl229 The `name` field is ultimately populated by a primary key up in the loader. So I don't think it's possible (in any realistic scenario) for it to not be defined.

PR #87, opened 2017-10-04
Attempt to fix 'server running webpack dev server'

oblodgett:
This is a breaking change that will cause the auto build servers to break... once the code is deployed gocd will need to be changed to get this working again. AGR-644


package.json
```diff
@@ -65,7 +65,7 @@
   "scripts": {
     "start": "webpack-dev-server --history-api-fallback --hot --inline",
     "prestart:docker": "npm install && npm run build",
-    "start-docker": "webpack-dev-server --history-api-fallback --host 0.0.0.0 --port 2992 --hot --inline",
+    "start-docker": "webpack-dev-server --history-api-fallback --host 0.0.0.0 --port 80",
```
ragingsquirrel3: this is still running webpack dev server.  Seems like it needs to run `npm build`.

PR #89, opened 2017-10-04
(for discussion) attempt unknown routes in WordPress prior to display page not found

sibyl229:
Hi there, 

The PR is created to enable endpoints created via WordPress available at the root path `/`, such as `/project`, `about-us` etc. @oblodgett requested this change for SEO reasons.

**Significant change is made to how routing is done with any route unknown to the javascript. If you have any concern, please let me know.** Thanks!

Technical detail:
- At the bottom of the Route definition in `routes.js`, any unmatched routes is handled by the WordPress component
- WordPress component attempts to retrieve the content from WordPress API, and upon failure or getting empty results displays `<NotFound>`


sibyl229:
Close for alternative approach of hardcoding urls instead of fallback.

PR #116, opened 2017-10-09
properly handling introns and other non-rendered types

nathandunn:
fixes AGR-702 and cleans up  lifting error


nathandunn:
https://agr-jira.atlassian.net/browse/AGR-702

I also fixed a linting error.  

This needs to be pushed back into develop at some point as well, but this fix will hopefully come into @pjhale 's widget.

PR #118, opened 2017-10-09
fixed some scaling and clumping issues

nathandunn:
Fixes scale shown for small and large ranges of genome features.


nathandunn:
https://agr-jira.atlassian.net/browse/AGR-710

PR #123, opened 2017-10-10
Agr 649 - UI Sitemap.xml  for production

LucyHut:
I created a python util (generate_sitemaps.py) that is called in the Makefile before the build. 
It generates the sitemap index and wordpress pages sitemap files under agr_ui/src/public 
when you run the "make" cmd -

For now I hardcoded wordpress pages but the ideal would be to make a WP REST API call to get the pages then create a dictionary that maps the pages to react routes ... I did not add wp posts – thinking on creating a different sitemap for posts (wordpress-posts-sitemap). 

Olin sugested we implement this using a javascript package - so maybe someone else can take a look at it. I personally do not have enough time to check it out before 1.0.


oblodgett:
As stated three times now, this will not work in the current production environment. Please use the following example:

https://www.npmjs.com/package/sitemap

To generate the sitemap.js file via javascript.

PR #126, opened 2017-10-12
NO MERGE - refactor to draw properly

nathandunn:
fixes AGR-674

I fixes this on the server.  Now we have the problem that we need to call the gene method internally it fails to retrieve the internal genes:

https://agr-apollo.berkeleybop.io/apollo/track/Mus%20musculus/All%20Genes/13:53466884..53473074.json?name=Msx2

with this error on the branch.  

screen shot 2017-10-11 at 5 43 03 pm

Only need one level of recursion, but not really sure how to get that with D3.  

Attn @paaatrick  @sibyl229  @ragingsquirrel3

### DIFF
--- a/src/components/genomeFeature/GenomeFeature.js
+++ b/src/components/genomeFeature/GenomeFeature.js
@@ -10,7 +10,26 @@ class GenomeFeature extends Component {
 
   constructor(props) {
     super(props);
-    this.MAX_ISOFORMS = 10;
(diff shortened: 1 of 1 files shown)


nathandunn:
and @pjhale  if you have any idea

nathandunn:
Looks like this might be the best route: https://stackoverflow.com/a/14390484/1739366  would have to flatten the data-structure instead of recurse over it.

PR #144, opened 2017-10-31
starting to handle feature types appropriately (AGR 732)

nathandunn:
Fixes: 

https://agr-jira.atlassian.net/browse/AGR-732


nathandunn:
FYI, @pjhale

src/containers/genePage/genomeFeatureViewer.js
```diff
@@ -79,45 +82,32 @@ class GenomeFeatureViewer extends Component {
...
-              loadState: 'loaded'
-              , loadedData: data
-            });
-            return data;
-          });
-        })
-        .catch(() => {
-          // console.log(error);
+    this.setState({loadState: 'loading'});
+    this.transcriptTypes = getTranscriptTypes();
```
sibyl229: A minor thing here. If you need to work with constants in a React component, you don't have to make it an attribute of the component and define it in lifecycle. (Things that are neither prop or state is used very sparingly in React.) Simply define it before your class definition should work.

nathandunn:
Thanks @sibyl229

sibyl229:
Thanks @nathandunn !

PR #152, opened 2017-11-22
Agr 757

bryon-czoch:


paaatrick:
Change looks fine, but shouldn't this be going into `develop` since AGR-757 is a 1.1 issue?

PR #156, opened 2017-11-30
Don't store API-translated field name when sorting

paaatrick:
This fixes some bugs I found while doing more testing of the recent disease components refactor on dev. The gist is that the redux store should always hold the sorting field name that the react-bootstrap-table is expecting. We should only translate that to the name that the API is expecting at the very last minute (i.e. before we fire off the fetch request) and that doesn't go in the redux store.

### DIFF
--- a/src/actions/disease.js
+++ b/src/actions/disease.js
@@ -42,6 +42,22 @@ export const fetchDiseaseFailure = function (error) {
 };
 
 export const fetchAssociations = function (id, page = 1, limit = 10, sortName = 'default', sortOrder = 'asc') {
+  // TODO: this is a hack because the API JSON field names don't line up with
+  // the URL query param names. So we rectify them here. It might be better to
+  // do it as part of the table column definition.
+  switch(sortName) {
+  case 'diseaseName':
+    sortName = 'disease';
+    break;
+
+  case 'disease_species':
+    sortName = 'species';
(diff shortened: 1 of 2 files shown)


src/containers/diseasePage/diseasePageAssociationsTable.js
```diff
@@ -24,25 +24,6 @@ import ExternalLink from '../../components/externalLink';
...
-
-  case 'disease_species':
-    return 'species';
-
-  case 'geneDocument':
-    return 'gene';
-
-  default:
-    return 'default';
-  }
```
sibyl229: Not sure if it's a problem, but the `default` clause isn't here anymore.

src/actions/disease.js
```diff
@@ -42,6 +42,22 @@ export const fetchDiseaseFailure = function (error) {
...
+    break;
+
+  case 'disease_species':
+    sortName = 'species';
+    break;
+
+  case 'geneDocument':
+    sortName = 'gene';
+    break;
+  }
```
sibyl229: AGR-759-orthology-appearance

paaatrick: Yeah, that was intentional. There's really no work to be done in the default case. Anything that's not handled in the other cases should just pass through.

PR #161, opened 2017-12-08
AGR-749 AGR-746 - browser back button doesn't work properly on News & Events page

paaatrick:
The cause of the bug was confusion about how data from wordpress was loaded into and organized within the redux store. The fix here is a general refactor of the wordpress-related code so that the redux store has separate fields for the current page or post and the list of all posts. Actions, reducers, and selectors were updated in accordance. This also allowed a simplification of the wordpress components so that each one only has a single responsibility. 

I also attempted to standardize the nomenclature in the code by replacing 'wp' with 'wordpress' and  slightly altered the styling of the list of news posts to be a little more inline with other pages.


bryon-czoch:
Invalid prop `postList` of type `object` supplied to `WordpressPostList`, expected `array`.
    in WordpressPostList (created by Connect(WordpressPostList))

paaatrick:
Good call. Warnings have been fixed.

PR #165, opened 2017-12-15
Agr 746 - Links on the 3 News articles all go to the same page

bryon-czoch:


paaatrick:
Not going to push this as a hotfix

PR #175, opened 2018-01-25
update to bootstrap 4.0.0-alpha.6

bryon-czoch:


paaatrick:
Is this complete? This seems to break a lot of layout.

Also, in general, I'm not sure how much bang for the buck we get out of just going to the latest alpha version?

bryon-czoch:
Oh you want bang for the buck?  From the UI meeting there seemed to be question to how much it would actually break.

paaatrick:
I'm going to just close this request because it's not ready to go yet.

By "upgrade bootstrap" I didn't mean simply pointing to a new version of the CSS file. It also needs to include going through all the existing code to address incompatibilities (like changed class names), making any other changes to ensure that existing layout and functionality doesn't break, assessing dependencies like react-bootstrap, etc.

PR #181, opened 2018-02-26
Agr 846 - disease table genetic entity info

bryon-czoch:


src/containers/diseasePage/diseasePageAssociationsTable.js
```diff
@@ -1,3 +1,6 @@
+/* disable-eslint */
```
sibyl229: It might not be the best idea to disable eslint entirely. If there are troublesome lines, maybe consider `// eslint-disable-next-line` or only disable eslint only for relevant sections of the code? https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

bryon-czoch: Disable all the things!...  this was just left over from debugging.

PR #208, opened 2018-04-16
AGR-808 add pagination totals to gene DAT table

bryon-czoch:


bryon-czoch:
Good idea, but let's create a refactor ticket for that and put it off.

PR #216, opened 2018-04-20
External link icon

felixgondwe:
- Added css rules to add external link icon


paaatrick:
Is it possible to make use of the existing ExternalLink component instead? It will apply the same icon as this, but it also causes the link to open in a new tab which is something we got a lot of feedback on previously.

PR #225, opened 2018-05-03
AGR-960 wrap long links

bryon-czoch:


paaatrick:
Hmm... not sure. Setting `word-wrap` on the `externalLink` class effects a lot of links that I'm not sure we want to touch. And it seems like too much padding around the definition links, especially given the discussion about the size of that section in general that we had in the yesterday's call.

I'm not really even I'm that offended by the long link as is currently on build.

bryon-czoch:
Now that I think about it more, is there any real value in dumping a long url like http://www.orpha.net/consor/cgi-bin/Disease_Search.php?lng=EN&data_id=17137&Disease_Disease_Search_diseaseGroup=rickets&Disease_Disease_Search_diseaseType=Pat&Disease%28s%29/group%20of%20diseases=Hereditary-hypophosphatemic-rickets-with-hypercalciuria&title=Hereditary-hypophosphatemic-rickets-with-hypercalciuria&search=Disease_Search_Simple

If anything it's just pushing things down more.

bryon-czoch:
These are just sources to the definition, could save a lot of space using superscripts.

paaatrick:
I think that's a really solid suggestion. I'm going to be on the Disease WG call next week talking about this info section and I can pitch that idea to them.

paaatrick:
Closing this one for now. I think it's likely we'll take it in a different direction.

PR #226, opened 2018-05-03
AGR-956 Only show one highlight for each 'super' field 

kevinschaper:
(aka hide duplicates from multi-fields)


src/reducers/searchParsers.js
```diff
@@ -4,6 +4,19 @@ const FILTER_ORDER = ['biotype', 'species'];
 import { makeFieldDisplayName } from '../lib/searchHelpers';
 import { NON_HIGHLIGHTED_FIELDS } from '../constants';
 
+export function flattenWithPrettyFieldNames(highlights) {
```
paaatrick: You don't need to export this function unless you foresee using it in another module.

PR #234, opened 2018-05-14
AGR-899 Upgrade to Bootstrap 4

paaatrick:
Summary:

* Upgrade to bootstrap 4 and fix layout problems caused by new class names
* Pull bootstrap (and dependencies) from npm instead of CDN
* Dump react-bootstrap in favor of reactstrap for interactive elements because it supports bootstrap 4
* Use Sass to process stylesheets so we can have a more customized bootstrap theme


bryon-czoch:
the navbar could use some padding

paaatrick:
Yeah, good catch. I noticed that later, too. I've got some changes on my page layout branch that will fix it. They're kind of intertwined with some other changes there, so I'm not sure how easy it would be to pull them into this branch. Basically what I'm saying is: I'll fix it later, trust me.

sibyl229:
Hi @paaatrick Changing major dependency is definitely hard work. Everything else looks great! I'm just wondering about the new colour for linked text. I find it slightly harder to read than before.

paaatrick:
Yeah, looking at it again it might be too low-contrast against the white background. Let me see what I can do about that.

paaatrick:
Okay @sibyl229 see what you think about the new link color.

sibyl229:
@paaatrick That's much better. Thanks!

PR #236, opened 2018-05-18
Agr 983 upgrade to react-router 4

sibyl229:
AGR-983
React router 2 isn't compatible with React 16. This PR upgrades react router to version 4 in preparation for React upgrade.

React Router 4 is a complete re-write. So a lot has changed since React router 2. Here is the upgrade guide: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/guides/migrating.md

A key difference is that the [react-router will be the source of truth for URL  instead of the redux store](https://redux.js.org/advanced/usage-with-react-router). You can still get the `location` and `history` easily through `<Route>` or `withRouter()`. I wonder if anyone has strong feeling about about this (@ragingsquirrel3 ?)

Code affected most by this is has to do with Search (@kevinschaper) and Google Analytics. Certain top level containers (GenePage, WordpressPost, DiseasePage) are slightly affected.


kevinschaper:
@sibyl229 I couldn't actually spot the problem in the code, but when I tried the branch out locally, it looks like the table version of there results aren't showing up.  I'd guess it's somewhere in the neighborhood of ResultsTable.js or MultiTable.js

sibyl229:
Hi @kevinschaper good catch! The table version of search is fixed. I made a mistake with the URL for the api. 
Btw, in React router 4, it seems that `to` prop of Link (when passed an object) no longer handles the `query` attribute: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md#to-object Just letting you know for future reference, since it's a pattern you have been using.
Thanks for testing it. If there is anything else, please let me know.

paaatrick:
The google analytics part seems to be working okay. Looks like there are a number of conflicts, probably changed by bringing in the `bootstrap-4` branch. I'll take a crack and resolving those.

PR #238, opened 2018-05-21
minor code cleanup (spacing) and fixes bad URL encoding

nathandunn:
fixes AGR-992


nathandunn:
Fixes https://agr-jira.atlassian.net/browse/AGR-992

src/containers/genePage/genomeFeatureViewer.js
```diff
@@ -45,7 +44,8 @@ class GenomeFeatureViewer extends Component {
     }
 
     for (let name in names) {
-      trackDataWithHighlight = trackDataWithHighlight + (name == 0 ? '?' : '&') + 'name=' + names[name];
+      let encodedName = names[name].replace('\|','%7C');
```
paaatrick: Perhaps [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) would be a more general solution here?

nathandunn: agreed

PR #244, opened 2018-05-25
AGR-811 Add in-page navigation to data pages

paaatrick:
I haven't heard any additional feedback, so let's do this thing! Let's make some heads explode!


oblodgett:
Where is the 🍿 when you need it?

nathandunn:
@paaatrick  This looks great overall.  I'm surprised that the links aren't readily going back to where they were selected from. 

Only other comment, which may be off-topic, would be that the sub-section footers (and headers) are much larger than needed, but maybe they denote sections (or maybe we could be more explicit with sections)?

nathandunn:
Would be great (from some very future iteration), if we could provide a summary of the data (e.g., there is no disease data), but I'm guessing that would be a PITA here.

paaatrick:
Sounds like your first and last comments are handled by AGR-996. I'm not sure what's involved in making that work, but I agree it would be a nice addition.

The `x-placement` warning is coming from reactstrap which was brought in with Bootstrap 4. I think that will be resolved when we finish the React 16 upgrade.

Data summary in the navigation panel is a great idea... for later. Feel free to open an issue about it.

PR #246, opened 2018-06-04
AGR-927 Add automated description to gene info section

paaatrick:


src/containers/genePage/modGeneDescription.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ExternalLink from '../../components/externalLink';
+
+const ModGeneDescription = ({description, url}) => (
+  <div>
+    {description && <div dangerouslySetInnerHTML={{__html: description}} />}
```
nathandunn: So, why is this being set this way?  Is that because you'll have HTML in the description text? 

Wondering out loud (not for this PR) if we should be returning a JSON standard that gets rendered internally.  For later, though.

paaatrick: Yes, some of the MOD-provided descriptions contain HTML markup.

paaatrick:
Thanks for taking a look. Feel free to let me know if you have general UI suggestions.

PR #248, opened 2018-06-06
AGR-901 Add table of alleles to gene page

paaatrick:

PR #249, opened 2018-06-06
Remove unneeded packages

paaatrick:
The error that @bryon-czoch mentioned in #248 was caused by a new version (1.2.0) of mocha-jsdom that requires a higher version of jsdom. Instead of upgrading jsdom I'm just removing it all because we don't actually need it.


bryon-czoch:
That works

PR #250, opened 2018-06-07
Use react-scrollspy instead of vanilla bootstrap scrollspy

paaatrick:
More testing indicated that using the vanilla bootstrap scrollspy implementation wasn't always reliable. I think it might have been due to a timing issue around when the bootstrap scrollspy was initialized vs when the subsection components were mounted. Regardless, I'm finding that using this react-specific package is more reliable.


bryon-czoch:
Search is broke

PR #254, opened 2018-06-12
AGR-983 Upgrade to React 16

paaatrick:


bryon-czoch:
Scroll spy is broke

paaatrick:
Not surprising. The changes from #250 aren't in this branch.

PR #256, opened 2018-06-19
Agr 766 - Add carousel

bryon-czoch:


src/components/carousel/index.js
```diff
@@ -0,0 +1,217 @@
...
+
+        </CarouselItem>
+      );
+    });
+
+    return (
+      <div>
+      <style>
+        {
+            `.custom-tag {
```
sibyl229: Does this way of declaring style potentially cause conflict? If so, maybe you could take the CSS module approach that's already widely used throughout this repo?

src/components/carousel/index.js
```diff
@@ -0,0 +1,217 @@
+/* eslint-disable */
```
sibyl229: Just a reminder of the eslint clause here.

src/components/carousel/index.js
```diff
@@ -0,0 +1,217 @@
...
+        </div>
+      </div>
+
+    </div>
+    );
+  }
+}
+
+
+export default Homepage;
```
paaatrick: Can you make the filename agree with the component name? It seems like `Homepage` is the right name since it's doing more than just the carousel.

src/containers/wordpress/wordpressPage.js
```diff
@@ -47,8 +48,9 @@ class WordpressPage extends Component {
     return (
       <div className={style.wordPressContainer}>
         <HeadMetaTags title={title} />
+        {slug === 'home' && <Carousel /> }
```
paaatrick: Since it looks like the home page is now entirely disconnected from wordpress it seems like it would be better to handle this by changing the react-router configuration instead of adding logic here.

PR #267, opened 2018-07-16
added an API to Downloads and changed name to Data

nathandunn:
Fixes https://agr-jira.atlassian.net/browse/AGR-1039

I would have rather had "Data" be Downloads in the download, but these nav menus don't seem to work that way.  I think this is fine, though (but am open to suggestions).

screen shot 2018-07-16 at 2 45 17 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Ensure you've installed [nodejs][1].
 * Clone your agr copy and checkout the development branch
 
 ```bash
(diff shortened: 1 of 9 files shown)


paaatrick:
Well let's just change it so that it works they way we want it to then!

nathandunn:
Thanks @paaatrick   This is awesome!

PR #269, opened 2018-07-17
Special case for swagger route

paaatrick:


nathandunn:
@paaatrick  I think it looks good.  I pushed up a little bit cleaner implementation, but feel free to revert if there is a good reason.    

When I had done this previously, I had sent it to `/api` and matched it on `/api` with the same redirect, but it got confused.   This seems to work, though.

paaatrick:
That doesn't actually work for me. I just get a page with a header & footer but no content and a message in the console that says:

`warning.js?688a:36 Warning: You tried to redirect to the same route you're currently on: "/api/swagger-ui"`

I think we need to use `window.location` so that it react-router _doesn't_ try to handle the location change.

nathandunn:
@paaatrick   That's fine.  Just pull out my change.  What you had worked for me, as well.   

Not sure why its coming up for me differently, but such is life.

PR #288, opened 2018-08-03
Upgrade packages

bryon-czoch:
https://snyk.io/vuln/npm:react-dom:20180802


sibyl229:
Do you want to also update react to the same version?

paaatrick:
> This attack is possible only only if the following two conditions are true:
> 
> * The application is being rendered to HTML using ReactDOMServer API, and
> * The app includes a user-supplied attribute name in an HTML tag.
> 

We're not doing either of those things. So, upgrading seems like a fine thing to do, but I don't see any value in making this a hotfix.

PR #291, opened 2018-08-03
update packages

bryon-czoch:


package.json
```diff
@@ -29,10 +29,10 @@
     "popper.js": "1.14.3",
     "prop-types": "^15.5.10",
     "query-string": "^5.1.1",
-    "react": "^16.4.0",
+    "react": "^16.4.1",
```
paaatrick: Any reason to not also go to 16.4.2 here too?

bryon-czoch: why not

PR #298, opened 2018-08-21
AGR-990 scroll to page top

bryon-czoch:


sibyl229:
`ScrollContext` and `ScrollToTop` aren't meant to work at the same time. You might consider having a conditional, such that `ScrollContext` _only_ when it’s chrome and `ScrollToTop` _only_ when it’s another browser. Just to be clear what kind of scroll restore behaviour is happening.

bryon-czoch:
ScrollContext isn't actually doing anything...  Chrome's built-in history is https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration

PR #300, opened 2018-08-29
initial implementation of  AGR-1160 - google structured data

nathandunn:
just a reference for this proposal:

https://agr-jira.atlassian.net/browse/AGR-1160

will close for now, but I find its easier to look at code.

Would need to replicate for the disease page if we go forward.

### DIFF
--- a/src/components/headMetaTags.js
+++ b/src/components/headMetaTags.js
@@ -4,9 +4,23 @@ import Helmet from 'react-helmet';
 
 class HeadMetaTags extends Component {
 
+
   render() {
 
     let title = `${this.props.title} | Alliance of Genome Resources`;
+    let data = this.props.data;
+    let schema = undefined;
+    if (data) {
+      schema = {
+        type: 'application/ld+json',
+        innerHTML: JSON.stringify({
+          '@link': 'http://www.alliancegenome.org/gene/'+data.primaryId,
+          '@type': 'DataSet',
+          identifier: data.primaryId,
+          name: data.symbol,
+        }),
+      };
+    }
 
     return (
       <div>
@@ -16,14 +30,17 @@ class HeadMetaTags extends Component {
           ]}
           title={title}
         />
+        {schema &&
+        <Helmet script={[schema]} />
+        }
       </div>
     );
   }
 }
 
 HeadMetaTags.propTypes = {
+  data: PropTypes.any,
   title: PropTypes.string
-
 };
 
 export default HeadMetaTags;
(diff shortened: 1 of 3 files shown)


nathandunn:
Just creating for a reference.

nathandunn:
To make this work, I would:
-  change the @link to “url” 
- [ ] make description match name or gene description (at least until we get a gene description in there)
- [ ] make keywords contain the species, gene name, and gene
- [ ] org name is some version of “The Alliance of Genome Resources"
- [ ] version (2.0)
- [ ] license (whatever we are using): https://creativecommons.org/licenses/by/4.0/ CC BY 4.0

The final product would include what Sarala has below:

```
  "@context": "http://schema.org",
  "@type": "Dataset",
  "name": "The name of the item",
  "description": "A description of the item",
  "url": "URL of the item",
  "keywords": "Keywords or tags used to describe this item",
  "includedInDataCatalog": "URL of the containing DataCatalog",
  "creator": {
    "@type": "Organization",
    "name": "Organization name"
  },
  "version": "Version Number",
  "license": "License names"
```

christabone:
Have we agreed upon a license for the Alliance?

nathandunn:
Stacia say it was the one I put in but we can clarify. 

Nathan

srengel:
this license: https://creativecommons.org/licenses/by/4.0/ CC BY 4.0

src/components/headMetaTags.js
```diff
@@ -16,14 +41,17 @@ class HeadMetaTags extends Component {
           ]}
           title={title}
         />
+        {schema &&
+        <Helmet script={[schema]} />
+        }
```
paaatrick: Do we need a separate `<Helmet>` component for this? Can't it just be included in the other one? React won't pass a prop to a component if the value is undefined.

nathandunn:
Just a quick FYI.  We'll be talking about this on Tuesday (meeting in calender, but ping me for times) to get feedback on the implementation.   I expect we'll make further changes, though, sometime next week and we'll have the final implementation for 2.0.  

@a8wright  I adopted this approach based on feedback from folks at identifiers.org and the google structured data page.  Open to different implementations, though. 

@paaatrick  Good catch. I added the fix you recommended.  It does seem to blow up if I don't handle the undefined case, but maybe you'd intended something else.  With regard to the allele case, I think we should definitely handle that case when we get to it, but point taken.   We should have another update and then we can merge it in with your suggested hierarchy, which sounds more scalable.

PR #307, opened 2018-09-11
add evidence code tooltips

bryon-czoch:


bryon-czoch:
Whoops, yeah that wasn't ready.  The text mocked was not meant to be the final solution.  Is the API work done?

PR #314, opened 2018-09-24
Upgrade ribbon 1.5.2

paaatrick:
Upgrading to the latest version of the ribbon. There's an issue with the evidence code filter box getting clipped by the horizontal scroll container, but we can work that out separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "url": "https://github.com/alliance-genome/agr_ui.git"
   },
   "dependencies": {
-    "@geneontology/ribbon": "^1.4.12",
+    "@geneontology/ribbon": "^1.5.1",
     "babel-polyfill": "^6.26.0",
     "bootstrap": "4.1.2",
     "d3-axis": "^1.0.8",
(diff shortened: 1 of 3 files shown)


nathandunn:
I think there is a problem with the CSS:

Ribbon by itself:

screen shot 2018-09-24 at 12 27 58 pm
screen shot 2018-09-24 at 12 27 56 pm


Ribbon in AGR:

screen shot 2018-09-24 at 12 22 10 pm

paaatrick:
Yeah, it's because the ribbon CSS is applying a fixed width and height to those elements, plus padding and a margin. On the alliance site, we have a global `box-sizing: border-box` CSS rule on all elements. But the ribbon seems to be assuming that `box-sizing` will be `content-box`. 

There's nothing we can really do on the alliance site to fix this. We can't just turn the `box-sizing` setting off because it would break bootstrap. 

The ribbon code could just set the padding and not have explicit width and heights. If you want those boxes to be the same size you could use flexboxes. Or if all else fails you could just override the `box-sizing` setting on the necessary elements.

PR #316, opened 2018-09-28
[DO NOT MERGE] Upgrade build tools

paaatrick:
The major upgrades here are going from webpack 1 to 4 and babel 6 to 7. That necessitated rewriting our webpack config files and .babelrc file. Packages with a peer dependency on webpack were also upgraded.

The only other slightly large change is using favicons-webpack-plugin to inject icon tags into index.html instead of having them hardcoded in the template file.

The rest of the changes are because the latest version of eslint got better at finding where we had incorrect indentation.


sibyl229:
Can we merge this in after 2.0 freeze? Just because the large number of changes. But it's great that you upgraded all the dependencies!

paaatrick:
Yeah, there really is no rush on this. So I think I agree to not rock the boat this close to the 2.0 release. I'll change the title to indicate not to merge it until the release-2.0 branch has been created.

PR #320, opened 2018-10-08
AGR-1138 Add comparison ribbon interface to expression section

paaatrick:
Mostly this is new code. It did touch some existing code to facilitate sharing code between this and the orthology section.

Also takes care of some issue that came in from early testing: AGR-1207, AGR-1206, AGR-1205, AGR-1204, AGR-1212.

### DIFF
--- a/src/lib/fetchData.js
+++ b/src/lib/fetchData.js
@@ -1,6 +1,6 @@
 import $ from 'jquery';
 
-const TIMEOUT = 5000;
+const TIMEOUT = 30000;
 
 export default function fetchData(_url, options={}) {
   let _type = options.type || 'GET';
--- a/src/lib/fetchData.js
+++ b/src/lib/fetchData.js
@@ -4,26 +4,10 @@ const TIMEOUT = 30000;
 
 export default function fetchData(_url, options={}) {
   let _type = options.type || 'GET';
-  return new Promise(function (resolve, reject) {
-    // *** DEPENDS ON GLOBAL $ because $ can abort ***
-    $.ajax({
-      url : _url,
-      type : _type,
-      dataType:'json',
-      timeout: TIMEOUT,
-      success: (data, textStatus, response) => {
-        if (response.status !== 200) {
-          let error = new Error('API Error');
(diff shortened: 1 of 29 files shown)


src/lib/fetchData.js
```diff
@@ -1,29 +1,13 @@
 import $ from 'jquery';
 
-const TIMEOUT = 5000;
+const TIMEOUT = 30000;
 
 export default function fetchData(_url, options={}) {
   let _type = options.type || 'GET';
-  return new Promise(function (resolve, reject) {
```
sibyl229: I'm a little worried about these changes to the fetchData return value. Is it necessary?

fetchData is used by many parts of the existing code. They might rely on the promise to resolve to certain value as defined by the current fetchData. Changing to the $.ajax might break things.

I imagine we might want to move away from jquery and use the `fetch API` for ajax calls in future. Exposing the return value of $.ajax might lead to code relying on specific functions available only in $.ajax's return type, hence making it harder to migrate from.

paaatrick: I totally understand what you're saying. 

The reason for this change is to support request cancellation. Fetching the expression annotations can be a little sluggish. If a user does a lot of filtering or paginating on the table there could be multiple requests in progress at the same time. Then there's a race condition with no guarantee of what order they'll come back. So I needed some way to cancel the in-progress request before the next one goes out. The `jqXHR` object that `.ajax` returns has a method to do that. It also is a Promise-like object so any `.then` and `.catch` calls that are currently chained to `fetchData` will continue to work.

I think we could benefit from request cancellation in other parts of the code. All the tables that do remote pagination could use it. The search autocomplete could really use it, too. But before we do that I agree we absolutely should convert from jquery to the Fetch API. Its cancellation mechanism seems to be a little more sophisticated, and I think it'll work well for us. So maybe rewriting `fetchData` to use the Fetch API is something we can do right at the start of the next release.

paaatrick:
Ah interesting! I will take a look at that. All the more reason to loop back to this as soon as possible.

PR #321, opened 2018-10-09
AGR-1196  Simplified field name prettification for GO & UBERON facets

kevinschaper:
@paaatrick Critique my loop please.


paaatrick:
It's a fine loop **_I GUESS_**.

PR #322, opened 2018-10-10
Agr 1148 physical interaction

sibyl229:


paaatrick:
The `DataLoader`, `DataProvider` pattern is very interesting! I was getting annoyed lately with the amount of boilerplate needed to fetch data and get it into the redux store, but this completely cuts all that out! I want to say that there are still advantages to using the redux store, but I'm going to have to think about that more. Definitely another thing to consider post-2.0.

sibyl229:
@paaatrick DataProvider can dispatch actions and subscribe to redux store state, instead of tracking the data in local state. So we can have the best of both worlds, less boilerplate and sharing states between different parts of the application. 
We might need to think about how to structure the redux state to achieve that post 2.0.

paaatrick:
Ahh yeah, great point!

PR #342, opened 2018-10-29
AGR-1278 Fix bug where first page view isn't logged in google analytics 

paaatrick:


sibyl229:
Ah, makes sense! Thanks for the fix. Merging.

PR #347, opened 2018-10-31
New Ribbon & resizing to fit in AGR page

lpalbou:


package.json
```diff
@@ -12,7 +12,7 @@
   },
   "private": true,
   "dependencies": {
-    "@geneontology/ribbon": "1.5.3",
+    "@geneontology/ribbon": "^1.5.7",
```
paaatrick: I feel a little uncomfortable about the compatible caret here. No disrespect, but this package has had a history of releasing backwards-incompatible changes as patch versions. I'd rather be more deliberate about using specific versions.

nathandunn: I agree.  I think that's what we had by default as well.   @paaatrick  Do you want to test this against the expression ribbon?

PR #348, opened 2018-10-31
move warning rule with others

bryon-czoch:


paaatrick:
I don't think this works. max-warnings is a command line flag, not a rule:

https://eslint.org/docs/user-guide/command-line-interface#--max-warnings

PR #350, opened 2018-11-09
AGR-947 Support for related data links between search results

kevinschaper:
With a real => function and a change so that ScrollToTop cares about search as well as pathname

### DIFF
--- a/src/components/scrollToTop/index.js
+++ b/src/components/scrollToTop/index.js
@@ -10,8 +10,8 @@ class ScrollToTop extends Component {
 
   componentDidUpdate(prevProps) {
     // compare pathname, which ignoring changes in URL hash
-    const currentPath = this.props.location && this.props.location.pathname && this.props.location.search;
-    const previousPath = prevProps.location && prevProps.location.pathname && prevProps.location.search;
+    const currentPath = this.props.location && (this.props.location.pathname + this.props.location.search);
+    const previousPath = prevProps.location && (prevProps.location.pathname + prevProps.location.search);
     if (currentPath !== previousPath) {
       window.scrollTo(0,0);
     }
--- a/src/components/scrollToTop/index.js
+++ b/src/components/scrollToTop/index.js
@@ -10,8 +10,8 @@ class ScrollToTop extends Component {
 
   componentDidUpdate(prevProps) {
     // compare pathname, which ignoring changes in URL hash
-    const currentPath = this.props.location && this.props.location.pathname;
-    const previousPath = prevProps.location && prevProps.location.pathname;
+    const currentPath = this.props.location && this.props.location.pathname && this.props.location.search;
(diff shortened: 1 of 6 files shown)


src/components/scrollToTop/index.js
```diff
@@ -10,8 +10,8 @@ class ScrollToTop extends Component {
 
   componentDidUpdate(prevProps) {
     // compare pathname, which ignoring changes in URL hash
-    const currentPath = this.props.location && this.props.location.pathname;
-    const previousPath = prevProps.location && prevProps.location.pathname;
+    const currentPath = this.props.location && this.props.location.pathname && this.props.location.search;
+    const previousPath = prevProps.location && prevProps.location.pathname && prevProps.location.search;
```
paaatrick: I think this should be concatenating `pathname` and `search' right? Not just doing a boolean operation to pick one of them?

src/containers/search/resultsList.js
```diff
@@ -51,13 +53,36 @@ class ResultsList extends Component {
     );
   }
 
+  renderRelatedData(d) {
+    if (!d.relatedData || d.relatedData.length === 0 || d.relatedData[0] === '') { return ''; }
+
+    const _links = d.relatedData.map( x => {
+      let queryParams = {category: x.category};
+      queryParams[x.targetField] = x.sourceName;
+      const categoryLabel = CATEGORIES.filter( cat => cat.name === x.category )[0].displayName;
```
paaatrick: If you just want the first match anyway: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

src/containers/search/resultsList.js
```diff
@@ -51,13 +53,36 @@ class ResultsList extends Component {
     );
   }
 
+  renderRelatedData(d) {
+    if (!d.relatedData || d.relatedData.length === 0 || d.relatedData[0] === '') { return ''; }
+
+    const _links = d.relatedData.map( x => {
+      let queryParams = {category: x.category};
+      queryParams[x.targetField] = x.sourceName;
```
paaatrick: I think you can just do 
```
const queryParams = {
  category: x.category,
  [x.targetField]: x.sourceName
};
```

src/containers/search/style.scss
```diff
@@ -83,6 +83,14 @@ td > .detailContainer {
...
 }
 
+.relatedDataLinks {
+  margin-left: 1em;
+}
+
+.relatedDataLinks ul {
+  list-style-type: none;
+}
+
```
paaatrick: SASS THIS!!
```
.relatedDataLinks {
  margin-left: 1em;

  ul {
    list-style-type: none;
  }
}
```
Also, isn't `.list-unstyled` already handling the `list-style-type` for you?

PR #355, opened 2018-12-12
Jsonld agr 1318 agr 1319 agr 1345

nathandunn:
fixes AGR-1345

### DIFF
--- a/src/components/headMetaTags.js
+++ b/src/components/headMetaTags.js
@@ -9,16 +9,48 @@ class HeadMetaTags extends Component {
 
     let title = `${this.props.title} | Alliance of Genome Resources`;
     let data = this.props.data;
-    let schema = undefined;
+    let schemas = [];
+
+
+    schemas.push({
+      type: 'application/ld+json',
+      innerHTML: JSON.stringify({
+        '@context': 'http://schema.org',
+        '@type': 'Organization',
+        'url': 'https://www.alliancegenome.org',
+        'logo': 'https://pbs.twimg.com/profile_images/751097942778351616/OW7j5m0Z_bigger.jpg',
+        'email': 'alliance-helpdesk@lists.stanford.edu'
+      })
+    });
+
+
+    schemas.push({
+      type: 'application/ld+json',
+      innerHTML: JSON.stringify({
+        '@context': 'http://schema.org',
+        '@type': 'WebSite',
+        'url': 'https://www.alliancegenome.org/',
+        'potentialAction': {
+          '@type': 'SearchAction',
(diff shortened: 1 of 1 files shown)


nathandunn:
https://agr-jira.atlassian.net/browse/AGR-1345

nathandunn:
Note @sierra-moxon  I put in a note about adding a resolver in there, but wasn't sure the library, etc., that should go with it.

nathandunn:
Agreed.  I forgot to mention that about the dates.  Yes date created versus date updated versus date published would be great if we have any of them. 

Nathan

sierra-moxon:
I think we have dateProduced.  dataModified makes sense (at this point) to be equal to dateProduced.  datePublished is probably a bit trickier -- should that be the release date?   Should we wait until the results of the email conversation before publishing the Stanford email address?

src/components/headMetaTags.js
```diff
@@ -9,16 +9,51 @@ class HeadMetaTags extends Component {
...
+
+
+    // general view
+    schemas.push({
+      type: 'application/ld+json',
+      innerHTML: JSON.stringify({
+        '@context': 'http://schema.org',
+        '@type': 'Organization',
+        'url': 'https://www.alliancegenome.org',
+        'logo': 'https://pbs.twimg.com/profile_images/751097942778351616/OW7j5m0Z_bigger.jpg',
```
paaatrick: Would probably be better to use an image that we host instead of our weirdly cropped twitter profile picture. Even something served from our wordpress server should be fairly easy to use instead.

src/components/headMetaTags.js
```diff
@@ -9,16 +9,51 @@ class HeadMetaTags extends Component {
...
+    // specify actions
+    schemas.push({
+      type: 'application/ld+json',
+      innerHTML: JSON.stringify({
+        '@context': 'http://schema.org',
+        '@type': 'WebSite',
+        'url': 'https://www.alliancegenome.org/',
+        'potentialAction': {
+          '@type': 'SearchAction',
+          'target': 'https://www.alliancegenome.org//search?q={term}',
```
paaatrick: Extra slash before search

nathandunn:
@sierra-moxon  I got the help-desk email from our public site.  

Do we have a place release date is stored?    I don't see it stored anywhere on the page.

sierra-moxon:
I think the discussion about email is going to possible change that email from our site to something else, but I am not sure.  We don't have dateReleased stored, no.

nathandunn:
Ah.  okay, well we can change it if we need to.

PR #357, opened 2019-01-03
uses component which automatically fixes the height - AGR-1146

nathandunn:
Last piece to fix this: 

https://agr-jira.atlassian.net/browse/AGR-1146

ATTN @sibyl229

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -20,3 +20,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Deprecated
 - removed panther data and loads and ui components
 - removed omim data and loads and but not ui components
+
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "d3-axis": "^1.0.8",
     "d3-scale": "^1.0.6",
     "d3-selection": "^1.1.0",
-    "genomefeaturecomponent": "^0.2.5",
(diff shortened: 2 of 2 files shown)


nathandunn:
@sibyl229  I leave it up to you, so long as we don't forget. 

My 2 cents is that its problem not introduced in the build.   If I run `npm test` it (continues) to work.    However, previously working builds seem to now fail.  

I'm going to re-run this: https://travis-ci.org/alliance-genome/agr_ui/builds/456076434 (which was previously working) and it now fails. 

Possibly we can go back further, but I think its a problem with a change in travis.

nathandunn:
But we can wait until after Patrick has taken a look as well.

sibyl229:
@nathandunn I don't think we will forget this, because I regularly visit the list of PRs. And I'm sure Patrick will see if before making a major release.

I will take a look at your previous PR and see if I can figure out what broke the build

nathandunn:
@sibyl229  Sounds good. 

If you go back and rebuild that was working in the develop branch:

https://travis-ci.org/alliance-genome/agr_ui/builds

You'll see that you'll be able to break any prior working build on the "develop" branch.   Why largely I think its a travis (or other) dependency issue.

PR #364, opened 2019-01-31
AGR-1390 Add release number to page header

paaatrick:
The release number will be pulled from an environment variable during the build process. If you're running the site using the webpack dev server or if the environment variable was unset while building it will display "Release [dev]" instead. You can test locally by running `RELEASE=2.1.0 npm run serve` and going to localhost:8080.


package.json
```diff
@@ -96,6 +97,8 @@
     "sitemap": "node bin/build_sitemap.js",
     "lint": "eslint src --max-warnings 0",
     "pretest": "npm run lint",
+    "preserve": "npm run build",
+    "serve": "live-server --no-browser --entry-file=index.html --proxy=/api:$API_URL/api dist/",
```
sibyl229: Can I confirm with you the `live-server` is only for development (ie. not a production server)? If so, when should we use this instead of the webpack-dev-server?

sibyl229: Oh, never mind. This is for serving the built code in `dist`.

paaatrick: Yeah it's to test a "production-ish" environment locally. It could probably use a better name.

PR #370, opened 2019-02-11
Remove extraneous bootstrap row

paaatrick:
This was causing the genome viewer to have the wrong margins around it


nathandunn:
@paaatrick  I'm sure the change is fine. 

However, I'm getting something different:

```
Module build failed (from ./node_modules/eslint-loader/index.js):
Error: Failed to load plugin flowtype: Cannot find module 'eslint-plugin-flowtype'
```

screen shot 2019-02-11 at 12 18 23 pm

. . . when doing `npm i` or `make` with node 8. 

npm start won't work either.    

I realize this is a separate error, but its untestable as-is.   Any ideas?

PR #375, opened 2019-02-14
Babel preset

sibyl229:
Use https://www.npmjs.com/package/babel-preset-react-app This preset is the one that comes with create-react-app. 

It includes the two presets that we previously used: @babel/preset-env and @babel/preset-react, ut also a few other stuff, such as supporting rest and spread operations.

I had to use `require` instead of `import` in webpack.config.babel.js because `import` is tripping over some syntax problem.


paaatrick:
Naming the file `webpack.config.babel.js` instead of just `webpack.config.js` would previously allow use to write that file using the ES6 import syntax through... some kind of babel loader magic. I'm not totally sure why this breaks that, but at the same time I'm not really married to using ES6 syntax in the config file -- a lot of people seem to get by just fine without out. But just to make it clear that we're not using (or, rather, attempting to use) that feature, maybe we should rename the file to just `webpack.config.js`?

sibyl229:
Hi @paaatrick 

The ES6 module import problem is caused by this webpack bug: https://github.com/webpack/webpack/issues/5960

"import" should be the only thing affected. You can still use other ES6 features. So I think it's more appropriate to keep it webpack.config.babel.js. ES6 module is always a little special compared to other ES6 features :sigh:

paaatrick:
Ohh interesting! Okay then.

sibyl229:
Thanks @paaatrick !

PR #376, opened 2019-02-15
Replace jquery for data fetching: AGR-1232

felixgondwe:
- Add fetch-api polyfill
- Add sanity test case for fetchData module
- Update ES6 to ES8 in eslint config file
- Update webpack entry object to handle whatwg-fetch polyfill
- Update README remote API_URL from *dev* to *build*


sibyl229:
@felixgondwe I like this!

paaatrick:
Changes for the abort and timeout look fine, but the test seems to be failing now.

PR #378, opened 2019-02-23
Table can end up outside of viewport: AGR-1218

felixgondwe:
- Add fixed maxHeight
- No known side effects


paaatrick:
I took a look at these changes. I like some it, but I have some questions about other parts of it. Could be a good thing to talk about at this week's meeting.

felixgondwe:
Thanks for checking that. Sounds great.

paaatrick:
After discussion in yesterday's call, I think we're going to pause on this for now. Just to keep things organized, I'm going to close this PR, but let's keep the branch around because we'll probably want to refer back to it later.

PR #384, opened 2019-03-05
Using the new Ribbon (1.9.2): appearances, multi entities, correcting…  (AGR-1055, AGR-958, AGR-1203, AGR-1325)

lpalbou:
… summaryRibbon

@paaatrick and/or @oblodgett could you check this PR please ? 

I had to adjust the summaryRibbon component to use the new data model (entities that each contain blocks, rather than blocks directly, therefore allowing for multiple entities).

I have added one parameter in the expression ribbon `onClick` method for your lazy loading, since the method now return both the selected entity and the block (before only block)

There are various ways to tweak the appearance (see style.css where I import the main sass file), but we can always adjust during this test phase or for the 2.2 release .

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1034,9 +1034,9 @@
       "integrity": "sha512-rLu3wcBWH4P5q1CGoSSH/i9hrXs7SlbRLkoq9IGuoPYNGQvDJ3pt/wmOM+XgYjIDRMVIdkUWt0RsfzF50JfnCw=="
     },
     "@geneontology/ribbon": {
-      "version": "1.5.7",
-      "resolved": "https://registry.npmjs.org/@geneontology/ribbon/-/ribbon-1.5.7.tgz",
-      "integrity": "sha512-SRUZmS1A6NfULr6cAZ+R8KHJcAsNyg883sDDEBLBmgWU29g+bb7yTkqhfzmwCRqRdrlEpmtiBckYpWOBWw8olw==",
+      "version": "1.9.4",
+      "resolved": "https://registry.npmjs.org/@geneontology/ribbon/-/ribbon-1.9.4.tgz",
+      "integrity": "sha512-BT4GR8o5UugDcI11lhulHuA+CJ+Bh5uCjLxXZa9Z9YElKijCZwCNsj7mt0zhkRnnqWXguGpzEBx2UPTfMvJGSQ==",
       "requires": {
         "amigo2-instance-data": "^2.5.12",
         "axios": "0.16.1",
@@ -1045,22 +1045,6 @@
         "react-icons": "^2.2.7",
         "react-popper": "^1.0.2",
         "react-transition-group": "^2.3.1"
-      },
-      "dependencies": {
-        "eslint-plugin-react": {
-          "version": "7.12.2",
-          "resolved": "https://registry.npmjs.org/eslint-plugin-react/-/eslint-plugin-react-7.12.2.tgz",
-          "integrity": "sha512-6F8uOJXOsWWWD3Mg8cvz4onsqEYp2LFZCFlgjaTAzbPLwqdwRCeK78unbuEaMXYPxq8paXCzTpoaDCwXBvC/gg==",
-          "requires": {
-            "array-includes": "^3.0.3",
-            "doctrine": "^2.1.0",
-            "has": "^1.0.3",
-            "jsx-ast-utils": "^2.0.1",
-            "object.fromentries": "^2.0.0",
-            "prop-types": "^15.6.2",
-            "resolve": "^1.9.0"
-          }
-        }
       }
     },
     "@iamstarkov/listr-update-renderer": {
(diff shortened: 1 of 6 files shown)


zfinannee:
@lpalbou   I added the ticket IDs that I think this PR relates to in the PR's title; this associates the commits with the JIRA tickets.   If I got any of them wrong, please update.   Thanks!

lpalbou:
@zfinannee this PR solves the following JIRA tickets:
* https://agr-jira.atlassian.net/browse/AGR-1055
* https://agr-jira.atlassian.net/browse/AGR-1203
* https://agr-jira.atlassian.net/browse/AGR-1325

Partially solved JIRA tickets:
* https://agr-jira.atlassian.net/browse/AGR-958 (we now have counts for both annotations & classes and it is shown on mouse over on a slim. but we haven't add any column to the table)

As well as tickets in the GitHub repo of the ribbon:
* https://github.com/geneontology/ribbon/issues/75
* https://github.com/geneontology/ribbon/issues/69
* https://github.com/geneontology/ribbon/issues/67

And a few requirements discussed during calls but not translated into tickets (e.g. having a table appearance more similar to what you have on the AGR site):
screen shot 2019-03-05 at 9 08 42 am

It doesn't solve however:
* https://agr-jira.atlassian.net/browse/AGR-1063

However, I don't know enough about the unit tests you are doing on the AGR site to solve the current travis issues of this PR. There is no reason why the unit tests would fail on a line declaring a SASS variable in a SASS file. @zfinannee could you assign someone that knows about your current unit tests  ? The PR in itself works without errors or warnings, I have checked it both locally and on a server.

paaatrick:
I think that the tests are in fact exposing an issue with the ribbon package itself. The error can be reproduced outside of the Alliance tests using a barebones React application. Like so:

- `$ npx create-react-app ribbon-test`
- `$ cd ribbon-test`
- `$ npm install @geneontology/ribbon`
- Add `import Ribbon from '@geneontology/ribbon';` to App.js
- `$ npm start`

You'll see the same error message about not being able to import Sass files.

lpalbou:
@paaatrick I don't get any SASS import error and the integration to AGR site does work locally without error or warning too (see above screenshot). We know the SASS files are included as the error states (npm test on this branch):
```
/Users/albou/go-dev/agr-site/agr_ui/node_modules/@geneontology/ribbon/lib/sass/_variables.scss:4
$ribbon-primary-color: #1849b4 !default;
                     ^

SyntaxError: Unexpected token :
```

where $ribbon-primary-color is indeed the name of a Ribbon SASS variable and I also double checked for the `--copy-files` of the build-react-component and it's here.

It's like the mocha test doesn't recognize a SASS variable, and indeed for both local src/ and dependencies, it's not supposed to check that at all:
```
      {
        test: /\.(css|scss)$/,
        exclude: /src/,
        use: 'null-loader'
      },
      {
        test: /\.(css|scss)$/,
        exclude: /node_modules/,
        use: 'null-loader'
      },
``

Any idea ?

**Comment by paaatrick** [comment]
The error is resulting from this line in dataHelpers.js:

```javascript
import variables from './sass/_variables.scss';
```

Hopefully I'm not stating the obvious when I point out that importing a Sass file is not a built-in part of Javascript. Using a set of webpack loaders (style-loader, css-loader, sass-loader) is what enables it. So you don't get the error when running the alliance site or your own demo page because those happen to be compiled using the necessary webpack configuration. But the alliance tests and the very simple react application I mentioned in my last comment don't use those webpack loaders -- hence the error. 

I've never attempted anything like this before so I'm not totally sure off the top of my head what the right solution to this is. I'm not sure how a component can dictate what webpack configuration is required to use it (if that's even possible, I'm not sure it's a good idea).

lpalbou:
Merge this first: https://github.com/alliance-genome/agr_ui/pull/386 to prevent testing of SASS files.

paaatrick:
I really don't like the solution proposed in #386. It's using a deprecated feature (https://nodejs.org/api/modules.html#modules_require_extensions) to hide the issue instead of really resolving it. But at the same time I don't want to keep holding this up forever. I'm willing to bring that change in for now, but can we agree to take a closer look at this once the 2.1 release is out?

More general comments/requests about these changes:

1. Can you just being the changes from #386 into this branch? That way we can see it all working together without having to do a bunch of extra merging.
2. Please remove all commented-out code
3. Please set $ribbon-strip-tile-border-radius to 0
4. Please move the Sass variable overrides from `style.scss` to `_theme.scss` so that they're with the bootstrap overrides
5. Is it possible to disable the grow effect when you hover over a tile?
6. Is it possible to disable this table label thing? It feels redundant with the section heading just above it.
image

paaatrick:
I'm approving #386 and this despite my concerns because Laurent-Philippe is out of the country and unable to make any more changes today and we can't hold up 2.1 testing any longer. I'll make cases for the thing I feel need to be changed or discussed.

lpalbou:
The design (especially for all & other tiles which are not exactly slims) was discussed and appreciated but some may not be aware of all of those changes: https://github.com/geneontology/ribbon/issues/67

But to remove the border-radius of tile is easy, just overwrite this SASS variable:
```
$ribbon-strip-tile-border-radius: 0px;
```

In the `geneOntologyComponent`, you can remove the table label by adding the `tableLabel` attribute:
```
                <Ribbon 
                  borderBottom
                  config={config}
                  entities={entities}
                  entityLabel={POSITION.NONE}
                  maxColor={[37, 152, 197]}
                  oddEvenColor={false}
                  showing={false}
                  subject={id}
                  tableLabel={''}
                />
```

Font color and size can also be adjusted by overwriting the following SASS variables:
```
$ribbon-table-headline-font-size : 2rem;
$ribbon-table-headline-color : black;

```

The grow hover on a tile is not customizable at this time but would be easy to change for the next release.

PS: I separated the PR https://github.com/alliance-genome/agr_ui/pull/386 from this one, in order for the checks of this PR to be able to pass, and also as they are unrelated issues (to be further discuss)

PR #386, opened 2019-03-08
Stop mocha from testing sass files (MERGE first, then https://github.com/alliance-genome/agr_ui/pull/384)

lpalbou:
This issue was blocking the Ribbon PR since the beginning of the week.

Mocha should not be testing SASS files as we don't currently have a real strategy yet to test styling features.

Please merge the Ribbon PR after merging this one and the tests should now pass.


paaatrick:
I don't think this is the best approach but approving anyway for the sake of time. See conversation in #384 for context.

PR #395, opened 2019-03-18
Ribbon go api

lpalbou:
Simple fix to fetch data from api.geneontology.org rather than the old default monarch
Should improve availability


paaatrick:
Something looks off with this commit list. Was `ribbon-go-api` branched off of `release-2.1.0` but this PR is going into `develop`?

lpalbou:
@paaatrick you're totally right, my apology. The PR is now directed towards release-2.1.0

PR #396, opened 2019-03-18
AGR-1522 IE11 fixes

paaatrick:
@sibyl229 apologies for undoing some of your recent work with this PR! I couldn't seem to get ES6 builtin functions (like `Object.assign`) working with the `babel-preset-react-app` preset. I kept getting errors about duplicate babel polyfill libraries being loaded. If you have other suggestions, please let me know or add commits to this branch.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,12 @@
 {
-    "presets": [
-      "react-app"
+  "presets": [
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry"
+      }
     ],
-    "plugins": ["react-hot-loader/babel"]
+    "@babel/preset-react"
+  ],
+  "plugins": ["react-hot-loader/babel"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1916,7 +1916,7 @@
         },
         "util": {
           "version": "0.10.3",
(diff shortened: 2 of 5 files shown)


sibyl229:
Hi @paaatrick sorry about that. Roll back is a good plan! 

I'm a little concerned about the changes in src/index.js. With the change it will always load polyfill regardless whether a browser need it or not. This penalizes modern browsers by slowing them down to load things the don't need.

Maybe the src/index.js just need something like this:

```
  require(['@babel/polyfill', 'custom-event-polyfill'], () => {
    main();
  });
```

But if this doesn't work. Feel free to merge your fix in. And we can sort out the rest next release.

paaatrick:
Yeah I initially tried to keep that in, but I still had no luck getting those ES6 builtins working on IE11. It wasn't immediately clear to me why, but I think it might have something to do with the fact that these newer versions of webpack and it's plugins and presets are becoming much more geared towards ES6 modules (`import`) as opposed to CommonJS (`require`). 

We should definitely take a closer look at this in 2.2. I'll open a new ticket for that. I think the right way to do it now is outlined here: https://itnext.io/lazy-loading-polyfills-4b85c4951e73

paaatrick:
Created case AGR-1529

sibyl229:
Thanks @paaatrick 👍 Let's revisit for 2.2

PR #399, opened 2019-03-20
fix jbrowse location

nathandunn:
when clicking on the JBrowse link it goes to the wrong place . . this makes it go to the right place


scottcain:
While I don’t have a problem with this PR, it just means that when I moved to production it gets undone

nathandunn:
@scottcain  / @sierra-moxon  it sounds like this should be closed?

PR #409, opened 2019-04-30
Fix AGR-1170

nathandunn:
fixes: https://agr-jira.atlassian.net/browse/AGR-1170


nathandunn:
fixes: https://agr-jira.atlassian.net/browse/AGR-1170

Also does some refactoring

nathandunn:
@paaatrick  I don't hate this idea.   The current architecture only has it in one place, which is what I used.  

I would propose that if everything else looks good to move this forward and if you can have multiple helmet tags we can then inject it via that mechanism.  I'm not sure if it would work that way, though.

paaatrick:
I'd prefer to not move forward with this. Even back when it was first put in place I didn't like the fact that the logic to build the metadata was so centralized (https://github.com/alliance-genome/agr_ui/pull/300#pullrequestreview-153165515). I don't want to continue that pattern and furthermore introduce potentially fragile code that infers data types from the presence of certain object fields. 

Also, to be clear, I'm not proposing multiple helmet tags. I'm just saying is that the logic for generating the data type-specific schema information should be kept in the components that manage that data.

nathandunn:
Ah, okay.  Moving the logic is definitely more do-able.    I'll see what I can do.

nathandunn:
@paaatrick  I'm looking at this more.   Really needs a top-level restructuring that is out of the scope of this and affects much more than the gene and disease pages.  DataPage is injected everywhere.    I could sub-class that, but it ends up being a Jenga maze. 

I would suggest that you put in the restructuring (roughly) how you want it and I can make this PR work within that framework.   I could see a service class or a few different sub-classes to make this work. 

Otherwise I would just push this through as-is, but I am more amenable to working within a grand scheme if you have the time / inclination to refactor.

nathandunn:
@paaatrick  Conversely if there is an example pattern to follow I would be happy to look at that as well.   I just think the initial header tag implementation wasn't particularly modular, so this is less moving backwards and more inheriting what is already there, but I am open to changing if there is a new template to follow.

paaatrick:
I will try to take a closer look later this week

paaatrick:
Have a look at #413. It keeps the gene-specific JSON-LD schema generation entirely within the gene page component and passes the result to `HeadMetaTags` which combines that with the generic JSON-LD schema. Adding JSON-LD to the disease (or any future data type) page would then simply involve generating the appropriate schama(s) in the disease page component and passing them to `HeadMetaTags`.

If that makes sense to you feel free to merge it into master and then work off of those changes.

nathandunn:
@paaatrick  Looks good.  I'll test and merge it soon.

PR #414, opened 2019-05-06
fixes case where synonyms do not exist  AGR-1585

nathandunn:
fixes https://agr-jira.atlassian.net/browse/AGR-1585?atlOrigin=eyJpIjoiZTc2OTIzMzg1YWQ0NDk4MzkzZmI4NzM0MjQwMzQ4OGQiLCJwIjoiaiJ9

### DIFF
--- a/src/containers/diseasePage/index.js
+++ b/src/containers/diseasePage/index.js
@@ -70,7 +70,10 @@ class DiseasePage extends Component {
 
     const title = disease.name || disease.id;
 
-    let keywords = ['disease', data.id, data.name, ...data.synonyms, data.definition];
+    let keywords = ['disease', data.id, data.name, data.definition];
+    if(data.synonyms){
+      keywords.push(data.synonyms);
+    }
 
     let definitions = [data.definition];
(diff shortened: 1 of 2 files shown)


src/containers/diseasePage/index.js
```diff
@@ -70,7 +70,10 @@ class DiseasePage extends Component {
 
     const title = disease.name || disease.id;
 
-    let keywords = ['disease', data.id, data.name, ...data.synonyms, data.definition];
+    let keywords = ['disease', data.id, data.name, data.definition];
+    if(data.synonyms){
+      keywords.push(data.synonyms);
```
paaatrick: This is going to put the put the synonym array itself into the array of keywords:
`['disease', 'DO:12345', 'foo', ['synonym 1', 'synonym 2']]`

Did you mean to use `.concat()` instead of `.push()`?

nathandunn: should be:

    keywords.push(...data.synonyms)

Fixed now.

paaatrick: That works too!

PR #428, opened 2019-05-29
AGR-1556 - update carousel with text and banner images

motenko:


paaatrick:
Whoa new background images looking fresh!

PR #434, opened 2019-05-30
AGR-1605 Use checkbox column filters where applicable

paaatrick:


paaatrick:
I agree there are pros and cons to using the Immutable type instead of a native array for local state. Previously I had been using a native array, and it turns out I created a bug because I was using methods that mutate the array directly (`push` and `splice`). So I wanted to experiment with using an Immutable type to make to clear (mostly to myself 😁) to not do that. But, yeah, it does mean consumers have to be aware of that. PropTypes help to some extent, but it's still a valid concern.

sibyl229:
@paaatrick I agree with you using immutable data structure to model **states**. State is an internal representation. **Props** is a little different, because it's an interface. An alternative might be to accept native array as props and convert them to immutable as needed?

Generally avoid mutating native array helps preventing many bugs. The spread syntax and the `slice` method produce new arrays and are generally safer to work with. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_array_literals

paaatrick:
Oh yeah, doing the array <-> immutable List transformation on the way into and out of the component is a really good idea. I'll look into that.

sibyl229:
@paaatrick thank you!

PR #447, opened 2019-06-11
Fix AGR-1617

lpalbou:


src/components/disease/diseaseComparisonRibbon.js
```diff
@@ -158,6 +159,7 @@ class DiseaseComparisonRibbon extends Component {
   }
 
   render(){
+    console.log('DCR: ', this.state);
```
paaatrick: Remove debugging please?

src/components/disease/diseaseComparisonRibbon.js
```diff
@@ -97,9 +97,10 @@ class DiseaseComparisonRibbon extends Component {
 
   handleOrthologyChange(selectedOrthologs) {
     const { dispatch, geneId } = this.props;
-    let geneIdList = this.getOrthologGeneIds(selectedOrthologs);
-    dispatch(fetchDiseaseSummary(geneId, geneIdList)).then(data => {
-      this.setState({ summary: {}});
+
+    let geneIdList = ['geneID=' + geneId].concat(this.getOrthologGeneIds(selectedOrthologs));
+    dispatch(fetchDiseaseSummary('*', geneIdList)).then(data => {
+      this.setState({ summary: {} });
```
paaatrick: This setState call is a little confusing since it looks like the `summary` field is also updated in the very next setState call. Can this one be removed?

src/components/disease/diseaseComparisonRibbon.js
lpalbou: Done

src/components/disease/diseaseComparisonRibbon.js
lpalbou: Done

PR #449, opened 2019-06-13
Fix AGR-1624

lpalbou:
@paaatrick for this kind of small PR when modifying a URL for which you probably don't know the syntax, I am unsure of the interest of asking for a review, but here it is.


lpalbou:
I also added the newTab=false in this PR

PR #450, opened 2019-06-13
Fix AGR-1643

lpalbou:
Something that I noticed before. The setState does not always merge correctly when we have a substructure that looks like the one already present, hence this brut force update to setState on the first summary level, since the subsequent merge does not replace items in the array.

Anyhow, we'll move away from setState possibly in profit of REACT hooks.


paaatrick:
That's very strange to me. I've never encountered a problem like that before in similar circumstances, and it would seem to go against how React is stated to work. It shouldn't matter what the substructure looks like since it should just be replaced altogether:

> This form of setState() is also asynchronous, and multiple calls during the same cycle may be batched together.
...
Subsequent calls will override values from previous calls in the same cycle

(from: https://reactjs.org/docs/react-component.html#setstate)

Regardless, if you've tested it and it seems to work we can go with it for 2.2. I still think there's something a little fishy about the data flow through the disease section, but that can be looked at more carefully in 2.3.

Also don't hold your breath on hooks, who knows when we'll get around to using them 😁

PR #451, opened 2019-06-14
Fix AGR-1616, AGR-1608

felixgondwe:
- Add missing columns, remove console log
- No known side effects


lpalbou:
Hi @felixgondwe , is it by design that a gene doesn't have any value in the "Genetic entity" column ?

Screen Shot 2019-06-17 at 8 26 26 AM

Also, could you double check with Sue or Yvonne if the source field should indicate the group (as it's currently doing) or if it should indicate some MOD ids, like the expression ribbon does (see ticket 1608).

Thanks !

PR #465, opened 2019-08-19
AGR-1662 Make Disease and Expression section ribbons consistent

paaatrick:


lpalbou:
looks ok but build.alliancegenome.org is not accessible. I'll do more tests later

shall we rename annotationTable -> expressionAnnotationTable (to be consistent with diseaseAnnotationTable) ?

lpalbou:
I am merging this PR but there are still two issues:

1. click on all stages show no data even if there should be some

2. the tooltip (and metadata hence coloring) of both "all expression in location" and "all expression in stages" are the same (same values eg /gene/RGD:3889#disease-associations : 132 classes and 251 annotations)

I will create a ticket on Jira for this.

PR #473, opened 2019-09-27
AGR-1696 AGR-1697 Update default state of disease comparison ribbon

paaatrick:
These changes:
* make the stringent ortholog set added for comparison by default
* display the table of all disease annotations by default
* update the ortholog picker for both expression and disease to be more oriented toward adding/removing sets of genes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12381,9 +12381,9 @@
       }
     },
     "reactstrap": {
-      "version": "7.1.0",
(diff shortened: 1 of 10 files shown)


paaatrick:
Good question. I reviewed their changelog (https://github.com/reactstrap/reactstrap/blob/master/CHANGELOG.md) and I don't think any of the breaking changes they list concern us.

sibyl229:
@paaatrick Thanks for checking!

PR #479, opened 2019-10-01
Go ribbon

lpalbou:
GO ribbon using the same data model & REACT GenericRibbon component.

Added a "show only experimental annotations" based on Biological Function group call.

### DIFF
--- a/src/components/geneOntologyRibbon/index.js
+++ b/src/components/geneOntologyRibbon/index.js
@@ -2,8 +2,6 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { POSITION, COLOR_BY } from '@geneontology/ribbon/lib/enums';
 import HorizontalScroll from '../horizontalScroll';
-// import NoData from '../noData';
-// import LoadingSpinner from '../loadingSpinner';
 
 import GenericRibbon from '@geneontology/ribbon/lib/components/GenericRibbon';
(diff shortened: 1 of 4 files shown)


src/components/geneOntologyRibbon/index.js
```diff
@@ -1,61 +1,475 @@
...
+
+  fetchAssociationData(subject, group) {
+    if(group == 'all') {
+      var groups = this.state.ribbon.categories.map(elt => {
+        return elt.id;
+      });
+      group = groups.join('&slim=');
+    }
+    let query = goApiUrl + 'bioentityset/slimmer/function?slim=' + group + '&subject=' + subject + '&rows=-1';
+    return axios.get(query);
```
paaatrick: `axios` is not a listed dependency of the agr_ui project. I don't like the idea of relying on dependencies of dependencies. Any reason to not use the `fetchData` function un `src/lib/fetchData.js`? 

I guess a bigger question is is there a reason that this is making API calls directly in the component, as opposed to going through redux?

lpalbou:
Yep sorry, I didn't meant to merge it yet, I wanted to review it but that's the GO meeting this week so had no time.

The GO ribbon however is using the GO API so it does require an http request. I could use the default JS library but I do prefer axios a lot if we can add it to the list of dependencies. Thanks

PR #497, opened 2019-10-11
AGR-1731

nathandunn:
fix for #1731


nathandunn:
https://agr-jira.atlassian.net/browse/AGR-1731

PR #516, opened 2019-10-18
AGR-1753 added sort tag to phenotypestatment list

prmani:
Seems like it was in the ui end. @paaatrick , can you please confirm? Thank you


paaatrick:
This is going to get fixed in the API layer

PR #522, opened 2019-10-21
Small ribbon fixes

paaatrick:


lpalbou:
Seems minor but how the RibbonGeneSubjectLabel affect the link or the rendering ?

paaatrick:
It does two things

1) It displays the link as "gene symbol (short species code)" instead of "[species icon] gene symbol". That was a curator request discussed on the UI call a while ago. The consensus was that the icons were too small to be legible at that size.

2) It uses a `react-router` `<Link>` component so that it doesn't do a full page load on navigation

lpalbou:
Well, there are several things here but:
* change of size: ok
* using the NoData component: ok
* using this new formatting of gene label in the go ribbon: not ok and not discussed in the BF group. we kind of wanted to propose the other ribbons to align to this styling which is much nicer.. and from memory Stacia also wanted us to make the page overall nicer

so can you remove the RibbonGeneSubjectLabel from the GO ribbon ? then we can merge this PR

PR #531, opened 2019-10-23
Fixing the "other" category association tabler

lpalbou:
I will have to fix the API summary query as the number displayed doesn't match. It will be fixed on the API side and not agr_ui


lpalbou:
Note: I have also created a more general API ticket to get and filter results in a better way and avoid so many data transformation on the client side

PR #537, opened 2019-10-29
Agr 1807

lpalbou:


lpalbou:
Simple fix to accept also other reference types

PR #541, opened 2019-11-04
AGR-1827 (Warning: GO ribbon fix needed before this should be merged) expression categories

sibyl229:
**Warning!!!** Merging this without the GO fix would break the UI

I un-commented the changes @paaatrick made on the default expression slim for yeast.


paaatrick:
Waiting for https://github.com/geneontology/ribbon/pull/106

PR #560, opened 2020-01-16
AGR-1891 Add allele summary info to page

paaatrick:
Description and Additional Information fields are intentionally blank for now because that data hasn't been loaded yet. 

Also includes some refactoring of the `PageNav` component to make it a little more flexible.


paaatrick:
Uh oh. uglifyjs is unhappy with something.

paaatrick:
The root cause was a faulty import statement (thanks IntelliJ!). While investigating I discovered that the webpack UglifyJsPlugin has been deprecated (https://github.com/webpack-contrib/uglifyjs-webpack-plugin) so I replaced it with TerserWebpackPlugin, which seems to be the recommended replacement. 

@sibyl229 this is ready for review now if you've got a minute!

PR #568, opened 2020-01-31
AGR-1919 download buttons restored

prmani:


paaatrick:
The test that failed is that one that connects to build, which may have been down at the time so I'm moving this one along.

PR #574, opened 2020-02-03
Apollo docker integration

nathandunn:


nathandunn:
This is needed because the variant service libraries don't like the space in the genome name, though they both point to the same place.

paaatrick:
Some eslint errors to resolve. I think a couple of them might be fixed if you merge master into this branch but there's also one about the `console` statement you're introducing.

paaatrick:
Also is it worth tying this into the `SPECIES` constant in `src/constants.js` instead of a one-off thing in the component?

nathandunn:
@paaatrick  thanks.  I thought I'd forked off of master, but must have been an older one.  

I'll relint and retest here in one second.

nathandunn:
@paaatrick  you had approved this, right?

paaatrick:
Shoot first ask questions later right? 🙃 

Anyway, yeah I think we landed in the right place here.

PR #576, opened 2020-02-07
Agr 1917 v1

nathandunn:
https://agr-jira.atlassian.net/browse/AGR-1851

https://agr-jira.atlassian.net/browse/AGR-1917


nathandunn:
FYI . . there is a hack for yeast in there, but not sure what to do about it

nathandunn:
it has to do with data

PR #583, opened 2020-02-10
apollo urls and jbrowse urls are now consistent

nathandunn:


nathandunn:
@paaatrick  it works.  I bumped the version to prevent another unrelated thing.

PR #589, opened 2020-02-12
Changed Dockerfile to support multistage build

oblodgett:


paaatrick:
Should I concerned about the UIPR/BuildAndTest failure?

PR #592, opened 2020-02-12
Fix AGR-1948 AGR-1952 AGR-1953

nathandunn:
fixes:

https://agr-jira.atlassian.net/browse/AGR-1953
https://agr-jira.atlassian.net/browse/AGR-1952
https://agr-jira.atlassian.net/browse/AGR-1948

### DIFF
--- a/src/containers/genePage/genomeFeatureWrapper.js
+++ b/src/containers/genePage/genomeFeatureWrapper.js
@@ -51,8 +51,14 @@ class GenomeFeatureWrapper extends Component {
     this.jbrowseUrl = externalJbrowseUrl;
   }
 
-  componentDidUpdate() {
+  componentDidMount() {
     this.loadGenomeFeature();
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.primaryId !== prevProps.primaryId) {
+      this.loadGenomeFeature();
+    }
   }
 
   getSpeciesString(species){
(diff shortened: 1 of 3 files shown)


src/containers/genePage/genomeFeatureWrapper.js
```diff
@@ -51,8 +51,7 @@ class GenomeFeatureWrapper extends Component {
     this.jbrowseUrl = externalJbrowseUrl;
   }
 
-
-  componentDidMount() {
+  componentDidUpdate() {
```
paaatrick: The typical pattern is to make a data fetch call unconditionally in `componentDidMount` and conditionally based on checking `prevProps` in `componentDidUpdate`. See for example: https://github.com/alliance-genome/agr_ui/blob/f492f633334244bc9bf780ff3de83d58656ecc2f/src/containers/diseasePage/index.js#L34
As it stands this is firing off about half a dozen duplicate requests as the page loads which isn't desirable.

nathandunn: I know.   Okay . . I'll give that a shot.

nathandunn: I pushed the fix for this

nathandunn:
> I'm not clear how these changes address point 2 in the description of AGR-1953 about adding the sequence viewer in the "Alleles and Variants" section.

Oh, add I see . . add it to the alleles section of the gene page . . . okay

paaatrick:
Yeah please talk to Sabrina if you're unclear on the requirements.

nathandunn:
1953 in here. FYI @sabrinatoro

image

nathandunn:
Everything is in there now . . wasn't clear where to put it, so I emulated what we did on the alleles page .. . 
ready for review / feedback

PR #605, opened 2020-02-20
close div when selecting another element

nathandunn:
fixes https://agr-jira.atlassian.net/browse/AGR-1983


nathandunn:
Never mind, not quite yet working

PR #615, opened 2020-02-28
AGR-2003 Handle transgenic alleles

paaatrick:


paaatrick:
The test failure is that test that tries to fetch a gene from build again 🤦‍♂

PR #617, opened 2020-03-03
integrating legend help text

nathandunn:


nathandunn:
@paaatrick  Can you take a look? 

Its kind of there, but I also get a lot of errors in the DOM.

image

paaatrick:
@nathandunn I did some repositioning and restyling to match the mockup.

One more thing on your side: Sabrina pointed out that in the legend "splice acceptor variant" has no color block next to it.
image

nathandunn:
@paaatrick  Awesome, that looks a lot better.  I will take a look at the splice acceptor variant

PR #619, opened 2020-03-03
round two of variant fixes

nathandunn:
https://agr-jira.atlassian.net/browse/AGR-2016

- [x] variant legend only showing up if variants present
- ~~should have an unknown (black?) and other (?) in the legend~~ (I checked and these were all good as AFAICT)
- [x] should flag 3_prime (versus three_prime) and 5_prime
- [x] when hovering over the things with the same number they both highlight: https://stage.alliancegenome.org/gene/FB:FBgn0002936#alleles-and-variants (small error)
- [x] remove console.log() from LegendService
- ~~make allele of genes unique~~ I think we agreed that this should be done later.

--- 

- ~~pair consequences and impact~~

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1620,9 +1620,9 @@
       }
     },
     "agr_genomefeaturecomponent": {
-      "version": "0.2.19",
-      "resolved": "https://registry.npmjs.org/agr_genomefeaturecomponent/-/agr_genomefeaturecomponent-0.2.19.tgz",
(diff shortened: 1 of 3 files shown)


nathandunn:
image

nathandunn:
@paaatrick  I think this is ready to review.

nathandunn:
note that this was fixed by: https://github.com/GMOD/GenomeFeatureComponent/pull/48

PR #627, opened 2020-04-16
AGR-2057 support for nested facet values

kevinschaper:


src/containers/search/filterSelector/singleFilterSelector.js
```diff
@@ -41,35 +40,33 @@ class SingleFilterSelector extends Component {
...
+    let _key = `fv.${this.props.name}.${value.name}`;
+    let nameNode;
+    if (this.props.name.match('species')) {
+      nameNode = <i>{value.displayName}</i>;
+    } else if (this.props.name === 'category') {
+      nameNode = <CategoryLabel category={value.name} />;
+    } else {
+      nameNode = <span>{value.displayName}</span>;
+    }
+    let dotNode = null;
```
paaatrick, line 57: `dotNode` seems to always be `null`. Can it be removed?

src/containers/search/filterSelector/style.scss
```diff
@@ -44,6 +44,14 @@
...
 }
 
+ul.filterList {
+  list-style-type: none;
+}
+
+ul.filterList li {
+  list-style-type: none;
+}
+
```
paaatrick, line 54: Do we need custom classes for this? Would Bootstrap's `list-unstyled` (https://getbootstrap.com/docs/4.4/content/typography/#unstyled) work?

kevinschaper: That's what I tried first, but sadly it doesn't indent

PR #632, opened 2020-04-30
AGR-1357 Show multiple data providers in disease source columns

paaatrick:


nathandunn:
@paaatrick  it looks good to me so long as it doesn't break anything

PR #633, opened 2020-05-01
AGR-2038 communicate allele selection from table to viewer

sibyl229:
The alleles table on gene page will pass the following props to the `VariantsSequenceViewer` component:
- `allelesSelected`
- `allelesVisible`
- `onAllelesSelect`

Clicking on a row selects the allele. Clicking again removes the selection of _that_ allele.

Multiple alleles can be selected.

@paaatrick @nathandunn @pjhale what do you think? If you like, we can also discuss this over a call.


nathandunn:
@sibyl229  I like embedding the viewer within the table . .  this looks great.  

The only one missing (at least as I understand Sabrina's spec) is `allelesHovered` to match the hovering, if possible)

paaatrick:
This looks good to me! I like the use of hooks; I keep meaning to talk about it on a UI call but I fully support using them over class-based components. 

If you need to support row hovering in addition to (or instead of?? I'm not totally clear on the requirements here) clicking I think you can use the `rowEvents` prop (https://react-bootstrap-table.github.io/react-bootstrap-table2/docs/table-props.html#rowevents-object), just in case you hadn't seen that already.

Anyway, this all looks fine to me. Feel free to merge it in whenever you all are ready.

nathandunn:
> If you need to support row hovering in addition to (or instead of?? I'm not totally clear on the requirements here) 

In addition to.  But I agree, I like the model here as well.

nathandunn:
@sibyl229  this is the hover for the boostrap table:

https://react-bootstrap-table.github.io/react-bootstrap-table2/storybook/index.html?selectedKind=Work%20on%20Rows&selectedStory=Row%20Event&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

nathandunn:
I **think** there is onMouseOut event not shown in the documentation or onMouseLeave.

sibyl229:
@paaatrick @nathandunn Thanks a lot for the pointers! I will take at them and explore the ideas in a different PR. The `allelesHovered` prop sounds good!

@paaatrick I love Hooks! Glad you support it! I think the linter needs to add an exhaustive dependency check (not sure what it is called) to avoid problems of forgetting to declare a dependency in a hook. I will look into this.

nathandunn:
@sibyl229  A different PR makes sense.  This looks good to me.

sibyl229:
@nathandunn Thanks! I will leave it up to you and @pjhale to merge :)

PR #634, opened 2020-05-01
bumped version to make it work properly

nathandunn:
bump in version for GenomeFeatureComponent fixes https://agr-jira.atlassian.net/browse/AGR-1989


nathandunn:
Ready for review.

PR #636, opened 2020-05-07
Send props to gfc

nathandunn:
this provides a hook into the GFC component as part of:

https://agr-jira.atlassian.net/browse/AGR-2038


nathandunn:
@sibyl229  Was going to implement for the variants page, but it looks like we only have these hooks on the allele page (which makes sense).  It is ready for review.

PR #641, opened 2020-05-12
AGR-2033 AGR-1352 Break long allele names in table cells

paaatrick:


paaatrick:
Ahh whoops! I accidentally also committed AGR-1352's changes to this branch!

PR #642, opened 2020-05-12
AGR-1963 - Add 'Privacy, Warranty, Licensing' sub-menu

motenko:


motenko:
Note: swapped 'About' and 'News' menus around for better alignment of text in the footer

PR #644, opened 2020-05-13
added set visible alleles initial fix

nathandunn:
fixes: https://agr-jira.atlassian.net/browse/AGR-1947


nathandunn:
This is ready for view

PR #650, opened 2020-05-27
AGR-1928: integrating the ribbon web component

lpalbou:
This PR allows the usage of the ribbon web component in both expression and disease ribbons. Aside from ticket 1928, this should take care of : AGR-1565, AGR-1253, AGR-1902. Once the GO ribbon is also integrated (later PR), this will also solve AGR-1471, AGR-1093 and AGR-2000.

**Notes**:
* ~~the “all annotations” selection on startup for disease is not yet working~~
* subjectFocus: the main gene was shown in bold in the previous ribbon, have to reintroduced this capability
* hideFirstSubject: this was used in the old ribbon to show/hide the subject label when only a single gene was shown
* empty cells have a cursor stating not-allowed but under the hood, if one click on that empty cell, an event is still triggered - will correct that at the ribbon level code

### DIFF
--- a/src/components/disease/diseaseComparisonRibbon.js
+++ b/src/components/disease/diseaseComparisonRibbon.js
@@ -139,7 +139,7 @@ class DiseaseComparisonRibbon extends Component {
         </div>
 
         <HorizontalScroll>
-          <div className='text-nowrap' style={{'width' : '1100px' }} >
+          <div className='text-nowrap'>
             {
               summary.loading ? <LoadingSpinner /> :
                 <wc-ribbon-strips 
--- a/src/components/disease/diseaseComparisonRibbon.js
+++ b/src/components/disease/diseaseComparisonRibbon.js
@@ -14,8 +14,7 @@ import HorizontalScroll from '../horizontalScroll';
 import { STRINGENCY_HIGH } from '../orthology/constants';
 import { getOrthologId } from '../orthology';
 
-import { GenericRibbon } from '@geneontology/ribbon';
-import { POSITION, COLOR_BY, SELECTION } from '@geneontology/ribbon/lib/enums';
+import { SELECTION, POSITION, COLOR_BY, FONT_STYLE } from '@geneontology/wc-ribbon-strips/dist/collection/globals/enums';
 import HelpPopup from '../helpPopup';
 import DiseaseControlsHelp from './diseaseControlsHelp';
 import ControlsContainer from '../controlsContainer';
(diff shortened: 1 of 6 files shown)


src/components/disease/diseaseComparisonRibbon.js
```diff
@@ -128,23 +139,35 @@ class DiseaseComparisonRibbon extends Component {
...
-              hideFirstSubjectLabel
-              itemClick={this.onDiseaseGroupClicked}
-              newTab={false}
-              selected={selectedBlock}
-              selectionMode={SELECTION.COLUMN}
-              subjectLabel={subject => <RibbonGeneSubjectLabel gene={subject} isFocusGene={subject.id === geneId} />}
-              subjectLabelPosition={POSITION.LEFT}
-              subjects={summary.data.subjects}
-            />
+          <div className='text-nowrap' style={{'width' : '1100px' }} >
```
paaatrick, line 142: Is setting a fixed width here necessary?

src/components/disease/diseaseComparisonRibbon.js
```diff
@@ -128,23 +139,35 @@ class DiseaseComparisonRibbon extends Component {
...
+                  color-by='0'
+                  data={JSON.stringify(summary.data)}
+                  group-clickable={false}
+                  group-open-new-tab={false}
+                  id='disease-ribbon'
+                  new-tab={false}
+                  selected='all'
+                  selection-mode='1'
+                  subject-base-url='/gene/'
+                  subject-open-new-tab={false}
```
paaatrick, line 156: Is it no longer possible to supply our own subject label component? My concern is that now these links will (presumably, correct me if I'm wrong) be rendered with plain `<a>` elements. Those will cause a full page reload and give a bad experience for our single page application.

src/components/disease/diseaseComparisonRibbon.js
lpalbou: Since you already have the HorizontalScroll component and text-nowrap class, no, but otherwise the ribbon takes the width of the parent component and could wrap

src/components/disease/diseaseComparisonRibbon.js
lpalbou: That is correct, those are traditional <a> links that will cause a full page reload. This is by design as we also want those links to open new tabs for other integrations.

However, I think there are two ways to solve that:
* web components do accept slots, so it's possible to pass your own html code instead but I'll have to investigate this a bit further to see if there is no issue to inject a REACT html code
* i could possibly also create another parameter to specify that these links should not reload the page and use something like https://gomakethings.com/how-to-update-a-url-without-reloading-the-page-using-vanilla-javascript/

In both cases, this will need a bit more try, so probably fixed more at the next release as my priority will be on the GO ribbon

paaatrick: The slots thing sounds interesting. I don't know much about it, but if that works, great. If that doesn't work why not just keep it simple and provide a parameter that's basically a click event handler that gets put on the `<a>` elements? Then the alliance code can `preventDefault` and do the necessary navigation work. 

Obviously the GO ribbon needs to get done, but to me this is the next highest priority thing -- higher than the other things listed in this PR.

lpalbou:
Also fixed https://agr-jira.atlassian.net/browse/AGR-1822

PR #655, opened 2020-06-01
Fix AGR-2201

lpalbou:


lpalbou:
Forgot a condition for that special group "all": https://github.com/geneontology/wc-ribbon/issues/22

PR #656, opened 2020-06-01
Update new ribbon style

paaatrick:


lpalbou:
I am not a big fan of the brownish color selection, but I suppose it's a personal taste.

Other than that, it's working. We can still revisit later if we want to improve some styling.

PR #660, opened 2020-06-03
AGR-2162 Molecular consequences section on Alleles page

sibyl229:


sibyl229:
Thanks for noticing that! I will look into it.

PR #662, opened 2020-06-05
fix for AGR-2212

nathandunn:
fix for https://agr-jira.atlassian.net/browse/AGR-2212


nathandunn:
going to ahead and merge this as its blocking curation testing

PR #663, opened 2020-06-08
AGR-1928: GO ribbon [DO NOT MERGE YET]

lpalbou:
This PR includes the web component of the ribbon table and the integration in agr_ui.

There is still a mix of old and new ribbon code which needs to be cleaned up.

In addition, there seems to be a small issue regarding the group-by functionality. Will check that during the day.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1295,6 +1295,11 @@
       "resolved": "https://registry.npmjs.org/@emotion/utils/-/utils-0.8.2.tgz",
       "integrity": "sha512-rLu3wcBWH4P5q1CGoSSH/i9hrXs7SlbRLkoq9IGuoPYNGQvDJ3pt/wmOM+XgYjIDRMVIdkUWt0RsfzF50JfnCw=="
     },
+    "@geneontology/curie-util-es5": {
+      "version": "1.2.4",
+      "resolved": "https://registry.npmjs.org/@geneontology/curie-util-es5/-/curie-util-es5-1.2.4.tgz",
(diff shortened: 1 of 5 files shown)


lpalbou:
The last commit has mostly fixed the GO table but i still need to rewire the following functionalities:

- [ ] filter experimental button does change the ribbon strips but does not yet affect the association table

- [ ] the two default filters are missing: (1) remove direct annotations to protein binding; (2) remove of annotations resulting from a cross-aspect mapping

- [ ] filter evidence codes: not currently in the table, would be better outside as this should affect both the table and the strips (old issue/comment)

- [ ] NOT qualifier not yet shown

- [ ] alphabetical ordering of annotations

- [ ] optional folding/unfolding of cells containing multiple values (eg > 5 references)

I will try to fix most of those by tomorrow, the one that could take a bit more time is the two default filters as they were coded on the client side / ribbon and they really need to be ported on the API side.

lpalbou:
Also fixed https://agr-jira.atlassian.net/browse/AGR-1093 with https://github.com/alliance-genome/agr_ui/pull/663/commits/1a61c194c6a0347d8df416e5bf4a0f9ecdf4df08

lpalbou:
@paaatrick this can be safely merged to master.

I will still have those items to finish in the testing phase:
* the two default filters are missing: (1) remove direct annotations to protein binding; (2) remove of annotations resulting from a cross-aspect mapping
* optional folding/unfolding of cells containing multiple values (eg > 5 references)

As for the _filter evidence codes: not currently in the table, would be better outside as this should affect both the table and the strips (old issue/comment)_ , this will be done with 3.2. The BF group initially planned to remove it from the table anyhow and to put it in an advanced panel.

oblodgett:
@paaatrick @lpalbou does this make it into the release branch? Or should I take them anyway?

lpalbou:
IMO yes as the remaining is more bug fixing. Maybe we can wait for Patrick opinion ?

PR #665, opened 2020-06-08
AGR-2162 VEP round 2

sibyl229:
basically moving things around and re-phrasing things as requested


sibyl229:
@paaatrick @nathandunn Thanks for checking!

PR #674, opened 2020-06-12
fixes variants showing up

nathandunn:


nathandunn:
This took a lot longer than it should have, but everything is there

PR #692, opened 2020-06-26
AGR-2284 fixed variant location in link

sibyl229:
My fault, but surprised that the linter didn't pick this one up.


paaatrick:
Yeah I'm also surprised that the `no-undef` rule didn't flag that. Maybe I'll take a closer look when we start messing around with the eslint config

PR #699, opened 2020-07-16
AGR-2153 Remove jquery dependency

paaatrick:
This is intentionally going into `master` to get on the 3.2.0 release path. We don’t want this in 3.1.1.

Bootstrap’s javascript relies on jquery (for now; their upcoming version 5 will not). However for the most part we use dynamic bootstrap components via their reactstrap wrappers, and reactstrap provides its own jquery-free javascript. So the change here is to drop the jquery dependency and remove the last few places where we relied on bootstrap’s javascript (via data- attributes) and replace them with reactstrap components and a bit of manual state management when necessary.


paaatrick:
🎉 🎉 🎉

PR #703, opened 2020-07-20
Fix AGR-2316

lpalbou:
This PR fixes AGR-2316 and a few other issues:

* The GO table should not display the "Gene" column as the GO ribbon is designed to handle selection of a single cell and not of a column. This column was simply removed
* I think someone has removed the HOC withRouter of the GO ribbon component as clicking on the gene link to the left of the ribbon strips was now triggering an error (history missing in props) and not rerouting the page to the gene of interest. I don't know what happened, but i re-added the withRouter and this is now working correctly
* updated the NPM package of both the ribbon strips and table and updated the CSS for the GO ribbon table so that the color of the links matches the color of other links on the Alliance pages

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -113,6 +113,6 @@ wc-ribbon-cell.ribbon__subject--cell--no-annotation {
 }
 
 .table__row__supercell__cell__link {
-  color: rgb(38, 129, 166)
+  color: $link-color
 }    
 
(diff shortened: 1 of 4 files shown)


src/style.scss
```diff
@@ -112,3 +112,7 @@ wc-ribbon-cell.ribbon__subject--cell--no-annotation {
   }
 }
 
+.table__row__supercell__cell__link {
+  color: rgb(38, 129, 166)
```
paaatrick, line 116: Could you use the `$link-color` theme variable instead so that we don't have hardcoded colors scattered about?

lpalbou: Done

PR #706, opened 2020-07-21
AGR-2327 AGR-2326 AGR-2325 AGR-2324 SARS-CoV-2 gene page fixes

paaatrick:
Apologies for dumping all these into one branch. When I started I thought there would be fewer changes.

- For AGR-2327 I added a new property to the `SPECIES` objects (`enableSingleCellExpressionAtlasLink`) which controls whether that link is presented on the gene page
- For AGR-2326 I just needed to move a margin CSS class to lower in the tree
- For AGR-2325, the "undefined" string was being produced by the ribbon web component (the species name "SARS-CoV-2" breaks assumptions it has about what species names look like). After talking with Chris Grove, we decided it was best to just do an end run around the immediate problem by 1) ensuring that the gene name label doesn't appear when you're not comparing orthologs (this is good because I believe this was the pre-3.1.0 behavior that may have regressed with the ribbon web component) and 2) disabling ortholog comparison for SARS-CoV-2 genes (this is also good because it wouldn't really make sense to do anyway). This required a bit of refactoring in the `OrthologPicker` component to hoist the state of the checkbox up to the parent components and also to pick up another new property (`enableOrthologComparison`) on the `SPECIES` constants so that SARS-CoV-2 doesn't appear in the species list.
- For AGR-2324, I applied the fix for AGR-2000 to both places where new data is fetched to ensure the focus gene is always in the `subjects` list.

### DIFF
--- a/src/components/OrthologPicker.js
+++ b/src/components/OrthologPicker.js
@@ -14,11 +14,12 @@ import {
   UncontrolledDropdown,
   UncontrolledTooltip
 } from 'reactstrap';
-import {SPECIES, TAXON_ORDER} from '../constants';
+import {SPECIES as ALL_SPECIES, TAXON_ORDER} from '../constants';
 import {
   compareAlphabeticalCaseInsensitive,
   compareBy,
   compareByFixedOrder,
+  getSpecies,
   makeId,
   orthologyMeetsStringency
 } from '../lib/utils';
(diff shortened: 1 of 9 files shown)


lpalbou:
I saw the modifications to the go ribbon, was a bit surprised and then read again your explanations. Sounds good to me.

I don't have time to run locally that PR and test functionalities andI didn't see some of my modifs that were merged to 3.1.1 in the diff which is a bit weird (eg calling ribbon.selectGroup), but I am guessing it's ok.

PR #707, opened 2020-07-22
added hooks for SARS

nathandunn:
FYI, added hooks to get SARS features in there. 

Note @scottcain  the lookup word into jbrowse (and subsequenly, hopefully apollo) I have is: `SARS-CoV-2`, but let me know if it changes.

### DIFF
--- a/src/containers/genePage/genomeFeatureWrapper.js
+++ b/src/containers/genePage/genomeFeatureWrapper.js
@@ -62,7 +62,7 @@ class GenomeFeatureWrapper extends Component {
 
   generateJBrowseLink() {
     const geneSymbolUrl = '&lookupSymbol=' + this.props.geneSymbol;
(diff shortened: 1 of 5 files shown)


src/containers/genePage/genomeFeatureWrapper.js
```diff
@@ -77,7 +77,8 @@ class GenomeFeatureWrapper extends Component {
   }
 
   getSpeciesString(species) {
-    return SPECIES.find(s => s.fullName === species).apolloName;
+    // short-name is for SARS only
+    return SPECIES.find(s => (s.fullName === species) || (s.shortName === species)).apolloName;
```
paaatrick, line 81: Doing this lookup by taxon ID is probably better than `fullName` or `shortName`. Not sure if this component has access to that info currently.

nathandunn:
@paaatrick  those changes are in there now

PR #718, opened 2020-08-08
fixes AGR-2360

nathandunn:
fixes https://agr-jira.atlassian.net/browse/AGR-2360

its a bit of a hack


paaatrick:
```
/workdir/agr_ui/src/containers/genePage/genomeFeatureWrapper.js
  20:9  error  'SPECIES' is defined but never used  no-unused-vars
```

nathandunn:
@paaatrick  hmm. . . . maybe we shouldn't be allowed to check in on errors?

paaatrick:
Probably a good idea. We already have husky and lint-staged installed as dev dependencies if you want to take a look at it.

PR #719, opened 2020-08-10
Lint commit check

nathandunn:
I added the lint check, but I upped the warnings allowed until . . at least until we have the new linting file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,10 +119,8 @@
   "pre-commit": [
     "lint"
   ],
-
   "husky": {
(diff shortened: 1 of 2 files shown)


paaatrick:
I don't think you can have two keys in the `hooks` configuration object with the same key; one will shadow the other. I also don't really like the idea of opening the door to checking in warnings, even if temporarily. I'm going to close this for now and I'll take a closer look when I make the eslint config changes later this week.

PR #721, opened 2020-08-14
AGR-2309 react query integration

paaatrick:


paaatrick:
@sibyl229 sorry for the enormous PR. please let me know if you have questions and check out the documentation i posted in slack

paaatrick:
That seems reasonable. I'll experiment with that a bit before I merge.

paaatrick:
I made that change and it definitely simplified some things. So thanks for that suggestion.

I pulled the `results` field of `resolvedData` out into a field just called `data` because sometimes that can be passed directly to tables (other times those results need to be transformed first). That adds a bit of redundancy, but I guess it's okay for now.

I also noticed that working with the `supplementalData` field of the API response can be a little clunky too. Maybe in the future the table query hook can provide nicer access to that.

sibyl229:
@paaatrick wow, very nice! Thanks for making those changes! It's gonna help me a lot.

PR #727, opened 2020-09-02
AGR-2349 AGR-2350 AGR-2351 AGR-2352 Display of disease NOT annotations

paaatrick:
In order to support the special formatting of NOT annotations a new component is added (`AssociationType`) that encapsulates the formatting logic. That component is then used in the `formatter` field of various column definitions, as well as in the new `filterFormatter` field. With this, the data cells and the filter checkbox labels can share formatting logic, while the original values from `distinctFieldValue` are still used in constructing the query param upon filtering. This eliminates custom logic in `buildTableQueryString`. The `filterFormatter` field is also used to ensure species names are italicized in the filter dropdown (previously this was accomplished by passing a class name). I also refactored the `DropdownCheckboxFilter` component a bit to make it stateless and give it better prop names.

The other infrastructural change was to add the ability to pass an object to the ribbon-related hooks which will be serialized into query params. This might be a useful pattern for other hooks, as well, in the future.


sibyl229 (approved):
Looks great!

PR #728, opened 2020-09-03
AGR-2303  Changed alteration category back to allele

kevinschaper:


paaatrick (approved):
This is definitely the type of stuff we should add unit testing around next time 😅

PR #729, opened 2020-09-04
AGR-2301 Alleles table on Gene Page

sibyl229:
Added category, hasPhenotype, hasDisease

TODO
- [ ] (postponed to future release) once variant RS ID is available, continue with the variant work


paaatrick (approved):
Looks good!

PR #732, opened 2020-09-10
AGR-2387 AGR-2403 Transgenic allele table bug fixing

paaatrick:
- Remove synonym column
- Fix overflowing in construct name column
- Adjust spacing of final rotated column headers to keep them fully visible at widest viewport


sibyl229 (approved):
Looks great! Sorry, those notification slips through the cracks. If I'm not reviewing the PRs, I probably didn't see them, just ping me on Slack.

PR #741, opened 2020-09-17
AGR-2271

pjhale:
Fixes AGR 2271 by displaying all variants regardless of pagination.  Uses a separate API call in useAllVariants that ignores the table pagination but uses the filters on the page.  Hopefully this code is consistent with the other react query code... I'm not all that familiar with it yet so this may not be the most concise way to do it.

### DIFF
--- a/src/containers/genePage/alleleTable.js
+++ b/src/containers/genePage/alleleTable.js
@@ -13,6 +13,7 @@ import RotatedHeaderCell from '../../components/dataTable/RotatedHeaderCell';
 import BooleanLinkCell from '../../components/dataTable/BooleanLinkCell';
 import VariantsSequenceViewer from './VariantsSequenceViewer';
 import useDataTableQuery from '../../hooks/useDataTableQuery';
+import useAllVariants from '../../hooks/useAllVariants';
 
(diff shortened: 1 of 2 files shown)


sibyl229:
@pjhale you might need to rebase your branch off latest commit on `master`. It's probably what's causing the test to fail.

paaatrick:
I merged `master` into this branch which should clear up the test run failure

paaatrick (commented):
I like the overall approach. Two comments/questions:

1. I know there's already enough confusion in the terminology around alleles and variants, but since the API endpoint called by the new hook is `/alleles` I wonder if the hook should be named `useAllAlleles` instead.
2. Did you take a look at the implementation of the `useAlleleVariants` hook? I wonder if the same pattern of calling `useDataTableQuery` (as opposed to calling `useQuery` directly) could be used here as well. In your case you could override properties of the passed-in `tableState` instead of `DEFAULT_TABLE_STATE`. Then you wouldn't have to worry about re-implementing `getFullUrl`.

pjhale:
On the first point I have no issues with renaming the hook to useAllAlleles.  On the second point I did try to use useDataTableQuery first but I was having issues passing the filters through that method without them getting overridden or changing the tableState unintentionally.

paaatrick:
Okay, fine for now then. I might go and take a second look later, time permitting.

PR #746, opened 2020-09-23
AGR-2420 Generate jbrowse link at render time

paaatrick:
Gotta do this the React way, not the OOP way!


sibyl229 (approved):
👍

nathandunn:
As long as the link still works after you’ve clicked through, it’s fine by me and easier to read.

PR #748, opened 2020-09-24
AGR-1846 Support for linked cross references in dataset results

kevinschaper:


src/containers/search/detailList.js
```diff
@@ -19,7 +20,9 @@ class DetailList extends Component {
       let valueNode;
       let value = d[field];
 
-      if (Array.isArray(value)) {
+      if (field === 'crossReferenceLinks' && value) {
+        valueNode = ( value.map(ref => <DataSourceLink key={ref.displayName + ref.name + ref.url} reference={ref} />).reduce((a,b) => [a,', ',b]));
```
paaatrick, line 24: I think you could use the `CommaSeparatedList` component here instead of the funky `reduce` thing

PR #750, opened 2020-09-25
AGR-2401 : fix label

lpalbou:
Make it clearer that the filter will not show only experimental annotations, but will show only functions/terms supported by at least one experimental annotation. Hence other annotations can be present:

https://agr-jira.atlassian.net/jira/software/c/projects/AGR/issues/AGR-2401?filter=myopenissues


paaatrick:
This looks good. I reverted the package-lock.json changes since there was no corresponding change to package.json. I think that can be caused by using an other version of `npm` -- see here for more detail: https://stackoverflow.com/questions/47638381/why-did-package-lock-json-change-the-integrity-hash-from-sha1-to-sha512

PR #752, opened 2020-09-25
AGR-2330 : add error handling when data can not be fetched or ID is missing

lpalbou:
Essentially, just use catch for fetch functions and add a state "error" to alter the rendering in case of issues.
https://agr-jira.atlassian.net/browse/AGR-2330

Tests:
	* HGNC with mapping
		/gene/HGNC:11998#function---go-annotations
		display correctly
	* HGNC without mapping
		/gene/HGNC:53833#function---go-annotations
		Show "No function available for that gene"
	* Regular gene
		/gene/ZFIN:ZDB-GENE-990415-270
		display correctly

### DIFF
--- a/src/components/geneOntologyRibbon/index.js
+++ b/src/components/geneOntologyRibbon/index.js
@@ -14,7 +14,7 @@ import LoadingSpinner from '../loadingSpinner';
 
 import { withRouter } from 'react-router';
 
-// import NoData from '../noData';
+import NoData from '../noData';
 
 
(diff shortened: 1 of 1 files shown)


paaatrick:
Looks like there are some linting warnings to clean up.

Also would it make sense to use the standard `NoData` component in the `renderError` function to maintain consistent style and messaging?

lpalbou:
I fixed the LINT and NoData

PR #757, opened 2020-09-29
Fixes AGR-2341

pjhale:
Added in an array prop that holds all of the locations for a specific gene.  This gets propagated down to both genome browser windows.  For examples of this look to HGNC:7082 and HGNC:750.

I'm not sure that I'm using the CommaSeperatedList component exactly correctly.  Things are working but I'm getting a warning about using keys in lists from that specific component.

### DIFF
--- a/src/containers/genePage/VariantsSequenceViewer.js
+++ b/src/containers/genePage/VariantsSequenceViewer.js
@@ -25,7 +25,7 @@ export function findFminFmax(genomeLocation, variants) {
   return { fmin, fmax };
 }
 
-const VariantsSequenceViewer = ({ alleles, allelesSelected, allelesVisible, tableState, gene, genomeLocation }) => {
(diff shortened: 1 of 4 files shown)


paaatrick:
You're using `CommaSeparatedList` correctly, but like any component that renders multiple children those children should have a `key` prop (https://reactjs.org/docs/lists-and-keys.html). In your case, the children are the `<span>` elements held in the `coordinates` array. Each of those `<span>`s needs a unique `key` prop added to it.

I'd also recommend using `.map` as a much more idiomatic way to generate `coordinates` from `genomeLocationList` (instead of `.forEach` and `.push`)

paaatrick:
Also, if this is supposed to be going into 3.2 can you please change the base branch to `release-3.2.0`?

paaatrick:
Is this ready to go? @pjhale is there any reason this is still a draft?

pjhale:
Yes it should be good to go once the checks pass.

paaatrick:
When I look at GOCD I can see that the checks ran and passed for the last check-in. I don't know why it's not getting reported back here.

PR #759, opened 2020-10-05
AGR-1846 Dataset search updates

kevinschaper:
Category renamed, data provider note added, color set, ID collection in place of Source + Cross References

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -21,16 +21,6 @@ export const CATEGORIES = [
     displayFields: ['name', 'synonyms', 'id', 'biotype'],
   },
   {
-    name: 'go',
-    displayName: 'Gene Ontology',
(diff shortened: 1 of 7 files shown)


src/constants.js
```diff
@@ -44,7 +39,12 @@ export const CATEGORIES = [
     name: 'model',
     displayName: 'Model',
     displayFields: ['id','synonyms'],
-  }
+  },
```
gildossantos, line 42: Order suggested in Slack #search by Leyla (and confirmed by Stacia) was to simply add dataset to the end of the existing list. So: 1. Gene 2. Gene Ontology 3. Disease 4. Allele 5. Model 6. HTP Dataset Index. I edited the file to match that.

paaatrick (approved):

PR #760, opened 2020-10-05
AGR-2461 Fix SARS-CoV-2 jbrowse links

paaatrick:


nathandunn:
Needless to say that this makes me sad to see.   I think its fine for now, but maybe we all can get together with @scottcain  and figure out a better naming solution as opposed to having 4 different names.

nathandunn (approved):

paaatrick:
I agree. That would be great.

scottcain:
@nathandunn that's funny because I commented on the jira ticket that it seemed like a good solution :-)

PR #762, opened 2020-10-07
AGR-2452 long sequence in attribute list

sibyl229:
In variant summary, render popup for long nucleotide change (>300 bases)


paaatrick:
Here's what the build is complaining about:
```
/workdir/agr_ui/src/containers/allelePage/Sequence.js
  10:3  warning  'renderLink' is missing in props validation  react/prop-types
```

sibyl229:
@paaatrick thanks for checking! I pushed a fix for that.

PR #764, opened 2020-10-16
fixed for AGR-2471

nathandunn:
fixes AGR-2471


nathandunn:
I think the variable should be part of state so that it properly gets the updated property.

paaatrick (approved):

PR #767, opened 2020-10-27
Added code for base images

oblodgett:


paaatrick:
Will `ALLIANCE_RELEASE` become an environment variable without the `ENV` command? Does `ARG` do that as well?

PR #773, opened 2020-12-04
Changed routes to get alliance mine link working

oblodgett:


kkarra (approved):

motenko:
From my testing of the fix (of appending a slash at the end of the url) on build does not work. I still get a page not found error, when testing with private browser window. You might want to try pointing to this url (note: appended with /begin.do):  https://build.alliancegenome.org/alliancemine/begin.do  ... For MouseMine, if I recall correctly, we have an apache setting that transforms 'www.mousemine.org' to 'www.mousemine.org/mousemine/begin.do' to accomplish the same thing.

motenko (changes requested):
From my testing of the fix (of appending a slash at the end of the url) on build does not work. I still get a page not found error, when testing with private browser window. You might want to try pointing to this url (note: appended with /begin.do):  https://build.alliancegenome.org/alliancemine/begin.do  ... For MouseMine, if I recall correctly, we have an apache setting that transforms 'www.mousemine.org' to 'www.mousemine.org/mousemine/begin.do' to accomplish the same thing.

oblodgett:
@motenko yes that would work, however we are proxying the requests and not redirecting the requests. So I think its still in the UI that is getting the request and not knowing what to do with it.

kkarra:
When you click on 'AllianceMine' link from 'Data and Tools' it goes to 
https://build.alliancegenome.org/alliancemine/  <-- which give pages not found

BUT, if you 'reload' the page, it works fine and goes to 
https://build.alliancegenome.org/alliancemine/begin.do  <-- this is the right URL to go to

I believe there is a bit of config in React (constants.js & routes.js) and a bit in ngnix
https://github.com/alliance-genome/agr_nginx_env/blob/8da53ffc15706eaed797c5743a3bca100e8325fd/build_webserver.conf#L69

Perhaps, we do need to add /begin.do to the line above? although not sure why reload would add it and work fine..?

Can someone please take a look?

 Thanks!

paaatrick:
It works on reload because it is not being handled in that case by the React SPA router. The change to `routes.js` which Olin made is the correct change to force the route `/alliancemine` to _not_ get handled by the React router and instead be interpreted as a normal request to the backend. 

It appears that the AllianceMine application knows how to to the automatic redirect from `/alliancemine` to `/alliancemine/begin.do`. So as far as I can tell I think this is good to go.

PR #775, opened 2021-01-13
AGR-2569 Use SpeciesName component for proper formatting everywhere

paaatrick:
The `SpeciesName` component not only italicizes, but also makes sure the capitalization doesn't get mucked with by other style rules


sibyl229 (approved):
This is good. A universal way of handling species name. Sorry I didn't see this earlier.

PR #776, opened 2021-01-19
Changed field name Expression to Cell/Tissue and facet name Anatomica…

adamgibs:
…l Expression to Tissues in HTP dataset

### DIFF
--- a/src/containers/search/filterSelector/filterSelector.js
+++ b/src/containers/search/filterSelector/filterSelector.js
@@ -37,9 +37,7 @@ class FilterSelectorComponent extends Component {
         isShowMore = true;
       }
 
-      if(d.name.toLowerCase() === 'anatomicalexpression' && this.props.activeCategory ==='dataset'){
-        d.displayName = 'Tissues';
-      }
+
       return <div key={`filter${d.name}`}><SingleFilterSelector {...d} isShowMore={isShowMore} queryParams={this.props.queryParams} /></div>;
     });
   }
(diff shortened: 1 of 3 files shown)


src/containers/search/filterSelector/filterSelector.js
```diff
@@ -36,6 +36,10 @@ class FilterSelectorComponent extends Component {
       if (d.name.toLowerCase().includes('species') || d.name === 'category') {
         isShowMore = true;
       }
+
+      if(d.name.toLowerCase() === 'anatomicalexpression' && this.props.activeCategory ==='dataset'){
+        d.displayName = 'Tissues';
+      }
       return <div key={`filter${d.name}`}><SingleFilterSelector {...d} isShowMore={isShowMore} queryParams={this.props.queryParams} /></div>;
```
kevinschaper, line 43: Hmm, you shouldn't have to catch and re-write the name here as well as in searchHelpers.js... ohh, I see it, the change in searchHelpers.js should be in makeFieldDisplayName rather than makeValueDisplayName, then it should already be changed by the time it reaches this part of the code.

adamgibs: Isn't this change necessary to change the facet name from Anatomical Expression -> Tissues, while the change in lines 39-41 in searchHelpers.js is necessary to change the field name from Expression -> Cell/Tissues? Or perhaps I've misunderstood the problem.

kevinschaper (approved):

PR #777, opened 2021-01-21
AGR-2571 Get download file types from API response

paaatrick:
@oblodgett I assume this is an acceptable use of the `fileExtension` field in the FMS API response, yes?


paaatrick:
@oblodgett does this look good? Is `stableGzipURL` preferred over `stableURL`?

sibyl229 (approved):

PR #779, opened 2021-01-22
AGR-2303 Change Allele category display name to Variant/Allele [DON'T MERGE YET!]

kevinschaper:
plus basic results display config

I had thought this was in, but it's been sitting around since October. I don't think variants are actually available yet, so I'm not totally sure if this PR should go in yet or not. 

@gildossantos Does it make more sense to just hold on to this, so that if 4.0 goes out without variants, the category label stays as just Allele?  There's an argument to be made that MOD variants are represented within the Alleles, so maybe this should go in?

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -32,7 +32,7 @@ export const CATEGORIES = [
   },
   {
     name: 'allele',
-    displayName: 'Allele',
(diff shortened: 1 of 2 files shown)


gildossantos:
We're going to start from scratch with this.

PR #780, opened 2021-01-28
Removed Unnecessary Load

pjhale:
Loading then highlighting was causing the genomeFeatureComponenent to 
not be ready for manipulation... and breaking the highlight.


pjhale:
@nathandunn Could you please check this.  I cant see any reason we needed this extra genome browser reload.  It seems to be the cause of the highlight not working.

nathandunn:
@pjhale  I think its fine.  I the cases to check would be bouncing it from one position to the next, refreshes, clicking to different parts of the page, and clicking to pages with different genes and/or variants.

PR #781, opened 2021-01-28
(WIP) AGR-2535 allele details page and Feature Viewer

sibyl229:
This is still waiting on some API changes. The UI code should be close to its ready state. It would be nice to get reviews from you.

@paaatrick two noteworthy changes, 1) adding the Allele Details page 2) making the table component automatically derive distinct filter values from supplementalData.

@nathandunn @pjhale I moved some logic out of `VariantsSequenceViewer` to its parent components, in order to reuse this component on the new Variant details page. The logic that got moved out involves, 1) calculate fmin, fmax 2) hide viewer while loading 3) display message if no variant is mapped.

If there is anything you are concerned about, please let me know.

### DIFF
--- a/src/components/dataPage/dataPage.js
+++ b/src/components/dataPage/dataPage.js
@@ -1,19 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ErrorBoundary from '../errorBoundary';
-import NotFound from '../../components/notFound';
 
 import style from './style.scss';
 
-const DataPage = ({children, isLoading, isError }) => {
+const DataPage = ({children}) => {
   return (
     <div className={style.dataPage}>
-      {
-        isLoading ? null : isError ? <NotFound /> :
-        <ErrorBoundary>
-          {children}
-        </ErrorBoundary>
-      }
+      <ErrorBoundary>
+        {children}
+      </ErrorBoundary>
     </div>
   );
 };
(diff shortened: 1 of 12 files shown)


nathandunn (commented):
I think moving the logic out makes sense. 

Just need to re-test it when the API changes come through.

sibyl229:
@nathandunn great! Let's do that. The API is mostly functional and the Feature Viewer is being rendered. One case that I haven't been able to check is the when fmax and fmin is outside the range of the gene, because the API is currently missing the variant coordinates for the new page.

paaatrick (approved):
This looks good to me. I think it definitely makes sense to have the DataTable component know as much about the structure of a "typical" API response as possible. If you have time to go back and look at the other table components that are manually calling `getDistinctFieldValue` just to make everything consistent that would be a nice bonus. Otherwise feel free to merge this whenever you're ready.

sibyl229:
@paaatrick thanks! About the manually invoked `getDistinctFieldValue`, I plan on doing those incrementally, as they require API side change - the filter naming in the URL parameter and the supplementData aren't guaranteed to be the same previously. But I made sure the change is backward compatible, that `filterable` manually set in the column configuration takes precedence over the auto-generated in the DataTable component.

sibyl229:
Since the code is functional, I'm pulling this in, even though the feature isn't complete. I have updated Sabrina on what isn't ready for testing.

PR #782, opened 2021-02-01
AGR-2396 added download for downloading tg allele table from gene page

prmani:


paaatrick:
Looks like there is some malformed syntax in TransgenicAlleleTable.js

PR #785, opened 2021-02-09
error handling for variant details page

sibyl229:
**fix missing import of NoData component**

An import statement was missing and somehow the linter didn't catch it.  Without the import statement, the UI would through an error when the API response contains no mapped variant.

**Separate error boundary for viewer and table**

So error in either of the component doesn't break another.


cmpich (approved):
Looks good!

sibyl229:
Thanks @cmpich !

PR #787, opened 2021-02-09
AGR-2124

adamgibs:
Implemented changes that were brought up in the UI WG meeting.

### DIFF
--- a/src/containers/search/filterSelector/excludeLink.js
+++ b/src/containers/search/filterSelector/excludeLink.js
@@ -1,0 +1,42 @@
+import {markAsExcluded, stringifyQuery} from '../../../lib/searchHelpers';
+import {Link} from 'react-router-dom';
+import React from 'react';
+import * as PropTypes from 'prop-types';
+
+/**
+ * @return {null}
+ */
+function ExcludeLink(props) {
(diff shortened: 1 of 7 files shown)


src/containers/search/filterSelector/excludeLink.js
```diff
@@ -0,0 +1,42 @@
...
+    return (
+      <span
+        style={{visibility: visibility}}
+        onMouseEnter={toggleStrikeThrough}
+        onMouseLeave={toggleStrikeThrough}
+      >
+        <Link
+          to={{pathname: SEARCH_PATH, search: stringifyQuery(newQueryObject)}}
+        >
+          <span title="Exclude term">&times;</span>
```
kevinschaper, line 23: It's probably worth doing an explicit tooltip here, the title from the span is a little slow to show up. I also noticed in the site we borrowed the idea from that they use a block level element, which seems like it gives a little bit bigger target area to click than just the x symbol itself.

kevinschaper (approved):

src/containers/search/filterSelector/excludeLink.js
adamgibs: Gotcha, that makes sense. Will do.

PR #795, opened 2021-02-18
sift and polyphen display in allele details table

sibyl229:
- color code sift and polyphen prediction and score
- add default formatter for column header checkbox
- remove empty value from distinct filter values

@cmpich the empty filter value might concern you, if you intended it to be a feature. But the API doesn't appear to work in that regard. ie. when siftPrediction filter is set to empty value, the response still include those with siftPrediction https://build.alliancegenome.org/api/gene/ZFIN:ZDB-GENE-001212-1/allele-variant-detail?filter.variantSift=&page=1&limit=100&sortBy= . I am omitting the empty values as possible filter values for now. It's likely not very interesting to know what variant has no sift prediction.

### DIFF
--- a/src/containers/GeneAlleleDetailsPage/GeneAlleleDetailsTable.js
+++ b/src/containers/GeneAlleleDetailsPage/GeneAlleleDetailsTable.js
@@ -15,6 +15,34 @@ import { getSingleGenomeLocation, findFminFmax } from '../../lib/utils';
 import VariantsSequenceViewer from '../genePage/VariantsSequenceViewer';
 import ErrorBoundary from '../../components/errorBoundary';
 
+const getSiftStyle = (sift) => {
+  switch (sift && sift.toLowerCase()) {
+  case 'tolerated':
+    return 'text-success';
+  case 'tolerated_low_confidence':
+    return 'text-success';
+  case 'deleterious_low_confidence':
+    return 'text-warning';
+  case 'deleterious':
+    return 'text-danger';
+  default:
+    return 'text-muted';
+  }
+};
+
+const getPolyphenStyle = (polyphen) => {
+  switch (polyphen && polyphen.toLowerCase()) {
+  case 'benign':
+    return 'text-success';
+  case 'possibly_damaging':
+    return 'text-warning';
+  case 'probably_damaging':
+    return 'text-danger';
+  default:
+    return 'text-muted';
+  }
+};
+
 const GeneAlleleDetailsTable = ({geneId}) => {
   const { isLoading: isLoadingGene, isError: isErrorGene, data: gene } = usePageLoadingQuery(`/api/gene/${geneId}`);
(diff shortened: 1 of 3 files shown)


src/containers/GeneAlleleDetailsPage/GeneAlleleDetailsTable.js
```diff
@@ -15,6 +15,34 @@ import { getSingleGenomeLocation, findFminFmax } from '../../lib/utils';
 import VariantsSequenceViewer from '../genePage/VariantsSequenceViewer';
 import ErrorBoundary from '../../components/errorBoundary';
 
+const getSiftStyle = (sift) => {
+  switch (sift && sift.toLowerCase()) {
+  case 'tolerated':
+    return 'text-success';
```
cmpich, line 21: The matching of predictions to a css is a bit weird. Rather than matching it to existing css classes that are used for success, warning and danger I would map it to new css classes with the color name or even better create classes such as 
text-sift-prediction-tolerated
text-sift-prediction-deleterious
etc.
That would be more explicit and does not tie any changes of text-danger, etc. with the colors of the predictions.

sibyl229: `text-danger`, `text-warning` are bootstraps classes designed exactly for this purpose - to give meaning to a colour, ie red means dangerous, orange means warning. This also maps quite well to the different levels of severity to SIFT and PolyPhen prediction.

This is how the design system works. Each colour in the colour palette is chosen with a particular meaning to allow consistency across the site, so that you can change the colour theme in one place that would be reflected across the site. It's regular practice and not weird at all.

sibyl229: In other words, the website should have a consistent way to communicate status to users through the consistent use of colour. What you suggest to avoid is actually desirable.

cmpich: What if we have more than three colors to use in our UI? If we are sure we only need danger, warning and success for the predictions then yes, we would map them to those three status flags.

sibyl229: We can have more colours if situation calls for it. I think we are good here right? Is there anything else for this PR before you check it off?

The goal with the colours really is to help the user understand information better. But simply having more colours do not necessarily help users understand what they mean. This is why a colour legend is often used in visualizations.

We associate special meaning to red, orange, green. We learned what they means through how we interact with other UIs and in our everyday lives. So as designers, we can use those colour to communicate information and be reasonable certain it will be understood. If we use other colours, we need to be very deliberate and ensure the meaning of it communicated.

cmpich (approved):
Looks good.

PR #796, opened 2021-02-22
AGR-1863 add rendering for the phenotypes column

sibyl229:
I assumed the column filter is called `filter.phenotypes`.


FB-ping:
you forget to change this line: // dataField: 'phenotypes', src/components/interaction/GeneGeneticInteractionDetailTable.js
should also you add a download link same as molecular interaction ?

sibyl229:
@FB-ping good catch! Fixed.

FB-ping:
test fine on local server

FB-ping (approved):
test fine on local server

PR #799, opened 2021-02-23
AGR-2535 allele details sorting and more filters

sibyl229:
@cmpich 

- [ ] Could you verify that I mapped the sorting labels to the correct sorting parameters you have? In the future, could you make sure those URL parameters (for sorting and filtering) match the language curators use. The use of different languages create a disconnect between the requirement from curators and the logic implemented in the code. This makes the code harder to reason with and maintain.

- [ ] Also, I notice filter.hgvsgName produces a 400 error. Here is an example: https://build.alliancegenome.org/api/gene/WB:WBGene00000913/allele-variant-detail?filter.hgvsgName=NC_003282.8%3Ag.424681G%3EA&page=1&limit=100&sortBy=

I am also waiting on the:
- [ ] filter for the location column
- [ ] the genomic coordinates of the

### DIFF
--- a/src/components/dataTable/AlleleCell.js
+++ b/src/components/dataTable/AlleleCell.js
@@ -4,7 +4,7 @@ import {Link} from 'react-router-dom';
 import AlleleSymbol from '../../containers/allelePage/AlleleSymbol';
 
 const AlleleCell = ({allele}) => {
-  return (
+  return allele.category === 'variant' ? allele.symbol : (
     <Link to={`/allele/${allele.id}`}>
       <AlleleSymbol allele={allele} wrap />
     </Link>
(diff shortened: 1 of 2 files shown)


src/components/dataTable/AlleleCell.js
```diff
@@ -4,7 +4,7 @@ import {Link} from 'react-router-dom';
 import AlleleSymbol from '../../containers/allelePage/AlleleSymbol';
 
 const AlleleCell = ({allele}) => {
-  return (
+  return allele.category === 'variant' ? allele.symbol : (
```
cmpich, line 7: I am not sure this is the right logic. Is this the first column? It should be allele.symbol in all cases. It will be populated by the API (even though, allele.symbol would be a bit confusing) transparently.

sibyl229: We don't have allele pages for HTP variants. It will link to a non-existent page that shows 404 if keep the link.

cmpich: Oh, now I see the logic. Looks good.

cmpich (approved):

sibyl229:
@cmpich great! thanks!

PR #803, opened 2021-02-26
AGR-2535 A series of small fixes for Allele/Variant Details

sibyl229:
- Bug in synonym display https://agr-jira.atlassian.net/browse/AGR-2622 
- Title of the details page: "Alleles and Variants details for [gene]"
    - Show ‘and’ instead of ‘or’
    - Instead of showing the gene in parentheses, show “for”
- The gene could/should link back to the gene page
- Between the sequence viewer and the table, add space or line, currently, it looks like this legend title is the header of the table.
- When there is no data, the details page should be suppressed
- When there is no data, the download on the “genomic variant” section on the Allele page should be suppressed


cmpich (approved):
Looks good.

sibyl229:
Thanks @cmpich !

PR #809, opened 2021-03-02
Fix AGR2626 and AGR2581 - for tests on master (another PR is for the release branch)

lpalbou:


lpalbou:
Since stage actually rebuild after a UI change, this PR is not needed.

PR #811, opened 2021-03-03
AGR-2535 allele details table update filters

sibyl229:


cmpich (commented):
Looks good.

sibyl229:
@cmpich I updated the `molecularConsequence` sorting option for both allele table and allele details table of a gene, as we discussed. Thanks for reviewing the first time round. I will merge this instead of bothering you again.

PR #812, opened 2021-03-04
AGR-2633 evidence code

sibyl229:
* multiple DOM nodes using the same ref
* tooltip flickering in and out with imprecise placement of cursor, by placing the tooltip to the side
* inline-block layout for multiple codes


adamgibs (approved):

sibyl229:
@adamgibs thanks!

PR #816, opened 2021-03-08
AGR-2643 allele detail notes

sibyl229:
@scottcain you are right that I just need to modify that same file. Please let me know if this look right to you. Thanks!


scottcain (approved):
Yep, this is all I was thinking you'd need.

PR #817, opened 2021-03-09
AGR-2637 alleleCategory filter name 

sibyl229:
@cmpich is the alleleCategory filter refer to the non-detailed allele table on gene page? Here is the updated filter name, if I understand it correct.


sibyl229:
@adamgibs could you help me out reviewing this PR? I don't think Christian is around to review PRs. And this just a name change for one of the table filters.

adamgibs (approved):

sibyl229:
Thanks @adamgibs !

PR #818, opened 2021-03-10
Support Development of shareable modules in the same repo

sibyl229:
**If there is anything I can do to make it easier for you to review the PR, please let me know.**

The objective of this PR is:
1) to **allow** development of shared and/or publishable modules (such as assets, theme and branding) in the same repo
2) to improve development tooling (in particular, Prettier, Storybook).
3) to support static pages (pre-rendered pages) with content from CMS

This might appear to be an intimidating PR, but the large number of changed files is a result of `git mv` the `src/` folder to `apps/main-app/src`. The effect is to turn the agr_ui repo from a single module repo into an [Nx monorepo](https://nx.dev/latest/react/core-concepts/why-monorepos)

The way to start a development server for agr_ui (main-app) is **exactly the same** as you do now. So is running linter and test.

I updated the README to explain the new folder structure and some new capabilities: https://github.com/alliance-genome/agr_ui/blob/modular/README.md

As part of this, I updated React to React 17, the [stepping stone release ](https://reactjs.org/blog/2020/10/20/react-v17.html) to make it easier to update React in the future.


nathandunn (approved):
@sibyl229  This is an impressive amount of changes, but I think the scope and goal is good. 

My only 2 questions are:

1. do you want to leave the `example-app` in there? 
2. are you going to publish individual packages (i.e., will there be a suite of helper libraries?).  Something like Lerna might make sense here, but perhaps that is redundant with what you already have.

sibyl229:
@nathandunn thanks a lot for the quick reply! Those are great questions.

1. The `example-app` current serves as a demo both for having multiple "apps" and for using the libraries in `libs/`. 
2. The `libs/` is intended for shared and/or published libraries. I currently have two libraries setup there `app-shell` for header and footer and `shared-assets` for images and stylesheets. The code currently in those libs are placeholders for what will eventually go there.
3. A nice detail regarding library publishing is that it isn't necessary to publish the library to use the it _within_ the repo. Only when used externally, such as in a different codebase, would it require the libraries to be published. So far, I haven't published those libraries, but they are publishable when needed. Simply by `cd` into to say `dist/libs/shared-assets/` and do `npm publish`.
4. It is similar to Lerna. I find Lerna to be more open-ended and it's up to the user to configure the tooling anyway they see fit. Nx also manages a monorepo, but its strength is in [the CLI](https://nx.dev/latest/react/getting-started/nx-cli), in particular code generation, code execution, and creation of dependency diagram for modules within the monorepo. Nx comes with tooling configured that are customizable for individual apps and libs. For example, I have customized the default webpack configuration by writing a function that takes the default webpack configuration (created by Nx for React) as an argument, and returns the final webpack configuration.

Please let me know if you have other questions.

adamgibs (approved):

nathandunn:
That you for the detailed explanation.  Really cool!

nathandunn (approved):

sibyl229:
Thanks a lot @nathandunn @adamgibs for reviewing!

I will give leave the PR open for @pjhale and @lpalbou to take a look. 

In the meanwhile, please feel free to file and merge as usual. And I will continue to merge `master` branch into this one, as I have done a few times. 

After the PR is merged, I will let you all know to check out new code to avoid merge conflicts. If you have exiting branches created before, I am happy to help resolve merge conflicts.

PR #819, opened 2021-03-10
fixes AGR-2646

nathandunn:


sibyl229:
@nathandunn does the viewer no longer care about the alleles visible for a particular page of the table? I recall the allele table has pagination and needs to inform the viewer what alleles are currently visible as a result of pagination.

nathandunn:
@sibyl229  That changed as part of this: https://agr-jira.atlassian.net/browse/AGR-2646 and https://agr-jira.atlassian.net/browse/AGR-2271 per.

nathandunn:
Cheers.  @pjhale  Let me know if you see any side affects.  I think we should be good and then we need to figure out a way to fix the rest of this in 4.1.

sibyl229:
@nathandunn You might want to create a PR against 4.0 branch if you want this on stage.

nathandunn:
@sibyl229  I just now realized that.  Doing it now.  Thanks.

sibyl229:
@nathandunn you also have a linter warning. Please fix before submitting a PR.

PR #821, opened 2021-03-10
Fix 2646 staging

nathandunn:


nathandunn:
PR for staging against: https://agr-jira.atlassian.net/browse/AGR-2646

PR #822, opened 2021-03-11
AGR-2649

adamgibs:
Changing the filter name to "Category" created a conflict with the higher-level "Category" name which resulted in the exclude link not working. I added a non-breaking unicode character to the name in order to differentiate between the two.


sibyl229 (approved):
🤣

PR #823, opened 2021-03-15
AGR-2652 turn cursor to normal for a row

sibyl229:
Fixing a mistake that changed the same of the cursor when shouldn't. Click on the table shoe does nothing.


adamgibs (approved):

sibyl229:
Great! Thanks @adamgibs for the quick response!

PR #825, opened 2021-03-17
Filtering Transcripts

pjhale:
On allele page filtering transcripts to only include those with 
consequences so that all can be seen.  Also updates widget version to 
0.3.11.

Making a draft until I can do a bit of testing to make sure we haven't introduced bugs.
Fixes bugs discussed in AGR-2483 and AGR-2216

### DIFF
--- a/src/containers/allelePage/AlleleSequenceView.js
+++ b/src/containers/allelePage/AlleleSequenceView.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import GenomeFeatureWrapper from '../genePage/genomeFeatureWrapper';
 import useAllAlleleVariants from '../../hooks/useAlleleVariants';
+import useDataTableQuery from '../../hooks/useDataTableQuery';
+
 
 
 function findFminFmax(genomeLocation, variants) {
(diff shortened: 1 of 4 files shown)


nathandunn (approved):
These looked good to me.  I did a little bit of testing, but not much and everything worked as desired.

src/containers/allelePage/AlleleSequenceView.js
```diff
@@ -22,6 +24,9 @@ function findFminFmax(genomeLocation, variants) {
 
 const AlleleSequenceView = ({ allele }) => {
   const variants = useAllAlleleVariants(allele.id);
+  let variantId = allele.variants[0].id;
+  const visibleTranscripts = useDataTableQuery(`/api/variant/${variantId}/transcripts`);
+  let isoformFilter = visibleTranscripts.data.map(a => a.id.split(":").pop());
```
nathandunn, line 29: linter says you should use single quotes instead of double

sibyl229 (approved):
@pjhale please fix the linter warning when you get a chance. 

I think the code might only work for alleles with a single variant, and it might be tricky to make this work with alleles with multiple variants depending on the APIs available.

I am not complete sure, but maybe the `/api/allele/{alleleID}/variants` endpoint instead of the `/api/variant/{variantID}/transcript` endpoint is more suitable. You should be able to retrieve the transcripts for all variants associated with the allele in the former endpoint. Here is an example: https://stage.alliancegenome.org/api/allele/WB:WBVar00091768/variants?page=1&limit=1000&sortBy=

PR #826, opened 2021-03-22
AGR-2671 molecular consequence filter

sibyl229:


cmpich (approved):
Looks good.

sibyl229:
@sibyl229 Thanks! I have another PR related to alleles, if you have could take a look at that one as well, it would be great.

PR #830, opened 2021-03-23
AGR-2674

pjhale:
Should fix the filtering issue.  This only updates the widget version.  The issue was the mouse id was different based on the API return and the apollo return.  In a future release the data should be looked at however the ID field from the API is equivalent to the name field from apollo.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1699,9 +1699,9 @@
       }
     },
     "agr_genomefeaturecomponent": {
-      "version": "0.3.9",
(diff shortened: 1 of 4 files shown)


pjhale:
@nathandunn I'm not sure that first commit should be on here.  It appears to have already been merged.

nathandunn (approved):

zfinannee:
I updated the name of this PR to be an exact string match to the id of the Jira ticket, so that this PR would associate to that ticket.   (If it isn't an exact string match, Jira doesn't recognize it...)

sibyl229 (approved):
@pjhale Looks good!

PR #831, opened 2021-03-23
AGR-2676 have viewer share same filter state as table but not pagination state

sibyl229:
For gene page and allele-details page for gene.

Forward to viewer via the `allelesVisible` prop up to 1000 alleles based on filtered but not paginated table state.


nathandunn (approved):
Wait for @pjhale  to have a look.  I was only able to take a cursory glance, but it looked good.

pjhale:
Looks good to me.  I'm able to filter and all of the variants I would expect to see are there including those on the next page.

pjhale (approved):

sibyl229:
@pjhale great! Thanks for checking!

PR #834, opened 2021-03-29
AGR-2679 AGR-2538 bugfixes

pjhale:
Fixes 2649.  Needs the latest browser version.  Also fixes an error that we get when an allele has an associated gene but no associated transcripts.


sibyl229 (approved):
@pjhale looks good to me!

nathandunn (approved):
👍

PR #840, opened 2021-05-05
AGR-2864 variant page

sibyl229:
Add the Variant Page. 

The API isn't fully ready - some fields are missing. The UI is developed to target the agreed upon response structure that the API would have for 4.1.

@pjhale Even though the necessary data for sequence feature viewer (variant page) development might not all be available through the API yet, this should provide a starting point (from the UI perspective). More details in src/containers/allelePage/VariantSequenceView.js and src/containers/allelePage/VariantPage.js .


cmpich (commented):
Looks good to me

sibyl229:
@cmpich thanks for reviewing! I will merge this in shortly.

@pjhale I added a minimal sequence feature viewer to the new Variant page. These new variant pages would mostly target the high-throughput variants, which don't currently show up in the viewer. In any case, I think it should be easier for you to continue with the sequence feature viewer work once this PR is merged in. So I will go ahead and do that.

PR #841, opened 2021-05-13
AGR-2667 release version number

sibyl229:
Use the FMS API to obtain the release version number, which is used to populate both 1) the release number in the website header and 2) the API call for the download page.

ALLIANCE_RELEASE environment variable is no longer in use and is retired.


adamgibs (approved):

sibyl229:
Thanks Adam for the speedy review! Olin told me later that the API endpoint for version number will change. So I will put this PR on holder for now.

PR #843, opened 2021-05-18
AGR-2957 updated variant search result link to go to variant page

adamgibs:


sibyl229 (approved):
Makes sense! Looks good!

PR #845, opened 2021-05-25
(Do NOT Merge) (Demo) remove bbop-manager-minerva and bbop-rest-manager as dependencies

sibyl229:
- Demostrated that removing dependencies `bbop-manager-minerva` and `bbop-rest-manager` would allow the CI to pass.


sibyl229:
The purpose of the demo is served. Closing :)

PR #846, opened 2021-05-28
Adding HTP Variant View

pjhale:
Adds browser version 0.3.14 that introduces an optional parameter in 
ISOFORM track views that renders a single SNP onto the viewer.

I've tested this on a few pages but its possible there are fringe cases that I haven't accounted for.  Assuming the variant object has a location parameter it should be passed onto the browser.  This does not make an additional call to apollo since we are currently only rendering the one variant.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1699,9 +1699,9 @@
       }
     },
     "agr_genomefeaturecomponent": {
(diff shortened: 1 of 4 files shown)


sibyl229 (approved):
Looks good!

PR #847, opened 2021-05-28
AGR-2665 hide allele/variant details link when there is no alleles/variants

sibyl229:
Hide allele/variant details link when there are no alleles or variants.

In the code, the alleles is used to refer to both alleles and variants, ie whatever is returned by the /api/gene/ID/alleles endpoint.


sibyl229:
Hi @pjhale here is another PR for you to review :)

PR #848, opened 2021-05-28
fetch only LTP variants for the sequence feature viewer

sibyl229:
API should be fixed before merging.


pjhale (approved):
Looks good... though I'm still not able to test on the build API.  Works fine using stage.

cmpich (approved):
Looks good.

PR #852, opened 2021-06-03
AGR-2708 : Alliance Pathways Section

lpalbou:
Ok, this is the clean version for the pathway widget.

Most of the bbop libraries were removed and it's now getting the data from a secured https API.

There will be changes in the coming months when GO refactors its own API to regroup routes that have been spread across the GO and GO-CAM APIs.


src/containers/genePage/index.js
```diff
@@ -127,6 +131,14 @@ const GenePage = ({geneId}) => {
           />
         </Subsection>
 
+        <Subsection help={<PathwayUserGuide />} t title={PATHWAY}>
```
sibyl229, line 134: the `t` prop in the Subsection component might have been a type?

sibyl229 (approved):
Thanks for tidying up the PR!

I added an inline comment where I think it might have been a typo.

Otherwise, feel free to merge it in, and we can address the potential consistency or usability concerns later.

src/containers/genePage/index.js
lpalbou: Yep it's a typo :) Thanks for spotting it

PR #857, opened 2021-06-03
(Do Not Merge) Demo call onAllelesSelect when button is clicked

sibyl229:


sibyl229:
This demo has served its purpose. Closing.

PR #858, opened 2021-06-03
Adding D3 and reactive code

pjhale:
Adds an onClick event that listens for clicks from the browser and can highlight rows there. This is not ready for merge as I need a new browser version that removes some of the old click-highlight code and I'd like to test on build but right now only stage seems to be working for the allele query.


sibyl229 (approved):
Thanks @pjhale ! The code looks good to me!

pjhale:
@sibyl229 Updated browser version and added catch to make sure code wont run when onAllelesSelect is not defined.  The browser change addressed a weird issue of some multi-allele variants getting arrays within arrays for which alleles they contained.  I still havent tested this with the build API but using stage everything works for me.

sibyl229:
@pjhale that makes sense! I will merge this, since this can be tested at /gene/WB:WBGene00000913/allele-details on build, and it satisfies this issue: https://agr-jira.atlassian.net/browse/AGR-2612 . Do you want to let Sabrina know that you have done it?

PR #859, opened 2021-06-04
AGR-2924 experimental condition in annotation popups

sibyl229:
This affects all phenotype and diseases tables that use annotation popups, and makes the experimental condition display available to them, if the data is present in the API response.


cmpich (approved):
Looks good.

PR #863, opened 2021-06-14
Added other endpoint for less data

oblodgett:


sibyl229 (approved):
Thanks for the PR! I made some adjustments because the same info reside under `releaseInfo` for the `/api/releaseInfo/summary`,  but under the root for `/api/releaseInfo`.

PR #865, opened 2021-06-16
More fixes for molecularConsequences

sibyl229:


src/containers/genePage/alleleTable.js
```diff
@@ -211,7 +216,7 @@ const AlleleTable = ({geneId}) => {
       style: {
         display: 'none',
       },
-      filterable: getDistinctFieldValue(resolvedData, 'filter.molecularConsequence'),
+      filterable: getDistinctFieldValue(resolvedData, 'filter.molecularConsequences'),
```
oblodgett, line 219: @sibyl229 @jt15 I think this line needs to get reverted otherwise the rest of this PR looks good

sibyl229:
@oblodgett @cmpich I reverted the filter to be `filter.molecularConsequence` (singular) for both endpoints:
- /api/gene/ID/alleles
- /api/gene/ID/allele-variant-detail

This was what was communicated to me, and at the moment the documented in the swagger UI. If this is inaccurate, please let me know.

PR #871, opened 2021-06-29
Removed Unneeded docker variables and added release date

oblodgett:
My hand at JavaScript

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,6 @@ ARG DOCKER_PULL_TAG=latest
 ARG REG=agrdocker
 FROM ${REG}/agr_base_linux_env:${DOCKER_PULL_TAG} as build-stage
 
-ARG ALLIANCE_RELEASE
-
 WORKDIR /workdir/agr_ui
 
 ADD . .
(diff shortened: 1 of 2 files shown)


sibyl229 (commented):
@oblodgett thanks for the PR. Everyone is welcome to contribute to this repo.

A gentle reminder that before you create a PR, you might want to check and address any ESLint warning if applicable. When you run the development server `npm start`, the ESLint warning would appear there and indicated "Compiled with warning".

Another thing I am uncertain about is that having both release (software version) and release date crowd up the header and breaks the mobile layout.  I wonder if maybe we should let the software number (4.1) go in favour of the release date. Seems that only internal members care about the software number?

Screen Shot 2021-06-29 at 2 27 40 PM

oblodgett:
Ok updated the lint and wrapped the text.

sibyl229:
@oblodgett thanks for the fix. The hamburger icon for the menu is still offscreen. 

I can make changes in the code, so you don't need to worry about that.

But it's important for me to know who cares about what information instead of filling up the header with all the information we have. Whether the "release ID" should be communicated to our users, when it essentially no longer represent a data release.

cmpich (approved):
Looks good.

PR #873, opened 2021-07-03
add js warning to index.html

bryon-czoch:
When JS is disabled in the end-users browser, AllianceGenome.org currently appears as if it's down.  :(


sibyl229 (approved):
Thanks @bryon-czoch ! Looks good!

PR #875, opened 2021-07-09
AGR-2228 AGR-2311 AGR-30587 Adding initialHighlight state and deletion support

pjhale:
Adding the 0.3.19 browser which adds support for showing multiple rows 
of deletions and setting an initial highlighted state of the browser to keep the view in sync with the table.
Most code changes in the genomeFeatureComponent repo.


sibyl229 (approved):
Looks good to me!

PR #878, opened 2021-07-20
Partial fix for gocam mouse over

lpalbou:
I was able to introduce a flexible strategy for the GO-CAM viewer. Currently the viewer will take focus:
* after a user stays more than 3s on the component (can be adjusted but 3s sounds a good balance)
* when a user click on the component

I have tried to hijack mouse events for the reactome pathway in different ways, but unfortunately wasn't able to prevent the default behavior. So this will require more discussion with reactome. However, double checking the current behavior, they don't automatically capture the mouse wheel either, they probably even require 3-5s to capture it, so it should not bother the casual scroll of the page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1319,9 +1319,9 @@
       }
     },
     "@geneontology/wc-gocam-viz": {
-      "version": "0.0.43",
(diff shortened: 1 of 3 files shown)


sibyl229 (approved):
Thanks for looking into the mouse capturing concern! The fix makes sense. I am happy if you merge it in.

PR #879, opened 2021-07-20
AGR-2708: make sure Reactome variable is also declared

lpalbou:
On very first load, I find that you might have an issue with the reactome library being loaded from the CDN. So I added an additional check if the Reactome variable is even declared (based on https://stackoverflow.com/a/138883/7608507).


sibyl229 (approved):
Looks good!

PR #880, opened 2021-07-20
AGR-2708: switch to first tab with data

lpalbou:
Pathway section currently has 3 tabs (pathway, reaction, GO-CAMs).

The requirement was to switch to the first tab with data (eg if no pathway, switch to GO-CAMs tab)


sibyl229 (approved):

lpalbou:
Note: example of gene without reactome data ZFIN:ZDB-GENE-990415-270

PR #881, opened 2021-07-22
AGR-3053 Fixing non-unique transcripts (Rat)

pjhale:
Added in code for both isoform only and isoform/variant to filter out non-unique transcripts in the widget.  This can be removed once we are sure the GFF and Apollo return is only unique transcripts.  We are filtering on the apollo field ID which in the case of rat is something like rna1234.  If the browser gets to a transcript with an ID that its already seen it will skip it.

No UI codebase changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1733,9 +1733,9 @@
       }
     },
     "agr_genomefeaturecomponent": {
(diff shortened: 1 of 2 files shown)


sibyl229 (approved):
Thanks!

PR #884, opened 2021-07-28
Improve shared header and footer

sibyl229:
Improve the header and footer in the `libs/app-shell` library.

- responsive menu (on smaller screens, user hamburger icon button to toggle the menu)
- social icons in the footer

To test, one way is to run storybook:

```
npm start app-shell:storybook
```

It will start storybook at http://localhost:4400/.

The same header and footer is applied to the MOD landing pages. So it's possible to test by running it. The storybook approach might be slightly easier.


adamgibs (approved):
Looks good! I tried it with Storybook and the responsive menu seems to work well on a smaller screen

sibyl229:
@adamgibs Thanks for reviewing!

PR #885, opened 2021-07-28
(Wait till MOD landing pages are ready) Add MOD landing pages links to header and footer

sibyl229:
This bit of work finished too early, before the MOD landing pages work is ready for the public :) I will leave this PR open, so that when the other work is ready, this can be pulled in without too much work on the agr-ui main app.

This PR applies the same shared header and footer already used in MOD landing pages to the main agr-ui app.

**Needs to wait until the MOD landing pages are ready for public before merge**, because the shared header/footer includes links to the MOD landing pages. I will make this PR **a draft to prevent accidental merge**, but I consider this ready. 

Links to the MOD landing pages use relative paths such as `/members/wormbase`. They are handled by the nginx **proxy** (https://github.com/alliance-genome/agr_ui/blob/master/nginx.conf#L10-L12), so will not work in local development environment.


adamgibs (approved):
Makes sense to me

PR #887, opened 2021-07-30
update reverse proxy for mod-landing-pages 

sibyl229:
To hide the Amplify domain name, use a custom domain for the mod-landing-pages.

Update the upstream server for mod-landing-pages (ie. members.alliancegenome.org). So that /members/[something] is proxied to the same path under the subdomain.

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -8,7 +8,7 @@ server {
 	}
 
 	location /members/ {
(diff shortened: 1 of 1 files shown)


adamgibs (approved):
Makes sense to me.

sibyl229:
Thanks @adamgibs !

PR #888, opened 2021-08-03
AGR-3081 Increase fetch timeout for allele/variants details endpoints

sibyl229:
- Allow fetch timeout to be overridden and specified per fetch request
- Add message warning about potential long wait time.


adamgibs (approved):
Looks good to me!

sibyl229:
@adamgibs Thanks for checking! There might be another related PR, because some might still want to disable the page to avoid overburdening the server.

PR #889, opened 2021-08-04
AGR-3081 allele buttons

sibyl229:
- display allele/variant details link below the summary table (instead of above)
- disable buttons when number of variants is large, and include a note to use the download page instead


adamgibs (approved):

sibyl229:
@adamgibs thanks a lot!

PR #900, opened 2021-11-04
Fix footer overflow caused by sitelinks

bryon-czoch:
only use top & left padding for teh sitelinks


bryon-czoch:
I realize now this isn't the problem, the footer styling needs a @media query

PR #901, opened 2021-11-05
feat(about): about page with static content - one page for each mod

valearna:
First implementation of about page for WormBase based on components without using template pages

https://agr-jira.atlassian.net/browse/SCRUM-688

This first implementation serves as a simple example on how MOD landing pages could be organized. I'm open to discussion on the structure of these pages @nuin


valearna:
After discussing the pros and cons of having separate mod pages vs. trying to create templates, @nuin and I decided to keep separate pages for each mod to have more flexibility for possible differences in the pages, so this PR is the starting point for the implementation

PR #906, opened 2021-11-12
First sprint sprint

nuin:
fixed members link
fixed media icons on footer
SGD log in front for decision
refactor scss for MOD landing
fixed title/logo bar size and padding


valearna (approved):
Everything looks good!

PR #910, opened 2021-11-22
Refactoring + tab label with MOD name

valearna:
- parametrized inner components and moved MOD specific content to content.js file
- created a Main mod landing page that feeds MOD specific content to the inner components
- refactored about and added new LinkToMOD component
- removed files from old structure
- renamed images in asset folder


azurebrd:
Using BEM convention for CSS, but I'm open to using whatever if people want.

valearna:
Thanks @azurebrd, I like how you organized the scss file and the idea of using BEM convention.

PR #919, opened 2021-11-30
TEST: Do Not Merge - adding mod landing page test workflow

dougli1sqrd:


ianlongden:
@dougli1sqrd, I made an attempt at getting this working. It was crashing in nx which I know nothing about and there were various complex solution on the net which I was unable to get going. Anyway, Instead I just ran npm run test and that seems to work. If this is bad then it is easily reversed but wanted to get this done before you take vacation.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  run-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0
+      - run: git fetch --no-tags --prune --depth=5 origin ${{ github.head_ref }}
```
dougli1sqrd, line 18: I think this line duplicates the first step, the actions/checkout@v2 ? Did you find you had to do both for this to work?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/setup-node@v2
+        name: setup node
+        with:
+          node-version: 12
+      - run: npm install
+      - run: npm run test
```
dougli1sqrd, line 24: oh weird that this just works without installing nx? The readme states that we have to isntall nx, which I thought was what my nx step was doing. But `npm run test` *does* seem to be running `nx test` and it works. So let's do it

.github/workflows/test.yaml
ianlongden: Yep, it might do, but some docs said it might have been just getting the commit rather than the branch etc.

.github/workflows/test.yaml
ianlongden: Yep, lets try it and see. Previously it did not work, so harm no foul. :-)

ianlongden (approved):

PR #924, opened 2021-12-02
JBrowse Link to Human Fix

pjhale:
Fixing link to human jbrowse within browser window and loading error on 
human browser window when variants are requested.

The issue described in KANBAN-33 was due to the "source" for human data being labeled as RGD.  We were building the JBrowse link based on this data.  It now looks at a track level variable to determine which genome to link.  This pull also fixes an issue with the human browser where it was requesting variants from apollo where none exist and failing to load.  Now it does not request these variants and loads fine.  

Also I don't really know who to request reviews for on these sorts of things.... so feel free to redirect this to others.


oblodgett:
Does this need to go in the release-4.2.0 branch? In order to get it into this release?

adamjohnwright (approved):

oblodgett (approved):

PR #925, opened 2021-12-03
Changed GO to GOC on footer for consistency with main page logos

nuin:
minimal change, merge if OK, otherwise delete branch

### DIFF
--- a/apps/main-app/src/constants.js
+++ b/apps/main-app/src/constants.js
@@ -127,7 +127,7 @@ export const NAV_MENU = [
       },
       {
         label: 'Gene Ontology Consortium',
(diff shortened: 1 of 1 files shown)


azurebrd (approved):

azurebrd:
@nuin merged and it's live on dev-public-ui, but FYI, the About still says GO (which might be right ?)  And the URL has members/go (which might be right ?)

nuin:
Yep, I will change the about later, thanks for letting me know. I think the URL is less important (many people don't care about it) but we can make it consistent.

azurebrd:
cool cool

PR #926, opened 2021-12-03
It is GOC everywhere now

nuin:
There's no more GO, just GOC


azurebrd (approved):

azurebrd:
Pulled to live, confirm https://dev-public-ui.alliancegenome.org/members/goc works and https://dev-public-ui.alliancegenome.org/members/go doesn't work

PR #927, opened 2021-12-03
GO banner added

nuin:
Original image was showing as a black stripe. I got just a part of it, but I will need to enhance it a bit to look good as the resolution is not really good.

Comments?


valearna:
The image looks good to me, apart from the very low resolution

valearna (approved):

azurebrd:
@nuin It won't help with the resolution, but you shouldn't need to crop/resize if you just change the "style.scss" .banner to have background-position: center center;  making it 
.banner {
  background-size: cover;
  background-position: center center;
  height: 200px;
}
That should also help with mobile, centering the image, letting the head of the rat show instead of just the back

azurebrd:
pulled and restarted dev-public-ui

nuin:
@azurebrd I just wanted to have the picture there. If all the pics had the same size, standard resolution,, etc, it would be easier for one solution for all. We have to work with what we have, that's life, but I will take a look at your suggestions

azurebrd:
@nuin Generally centering the pictures is probably a good thing, most of them should have something more relevant towards the middle vertically + horizontally, and that would make it so that in tiny screen like mobile, we get the middle of the picture, not the upper left.  If you're not opposed to it, I'll add that  background-position: center center;  on some PR.

PR #930, opened 2021-12-05
Resources

nuin:
All resources added, in the spirit of scrum, they are there, doesn't mean that they look good


azurebrd (dismissed):

azurebrd (approved):
I kinda don't think it should all start with a <br /> and that it should be part of the container, if it needs an extra <br />, but looks good.

nuin:
Yes, I agree, kinda of a hack, but it works and it’s what the ticket says😅

azurebrd:
live now

PR #931, opened 2021-12-06
reverted banner to previous stage

nuin:
Banner settings are reverted

### DIFF
--- a/apps/main-app/src/containers/modLanding/style.scss
+++ b/apps/main-app/src/containers/modLanding/style.scss
@@ -51,7 +51,7 @@
   padding: 0;
   margin: 0;
   width: 100%;
-  height: 370px;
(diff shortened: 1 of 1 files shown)


azurebrd (approved):

azurebrd:
Pulled to live.  I see what you mean about the RGD fit now @nuin .  I guess there could be some kind of RGD-specific css to make the vertical be higher than center to get more of the head, but yeah, this isn't so good with a wide screen.  Hopefully when we have specs RGD can give us something better.

nuin:
Thanks @azurebrd. I guess it’s the master and PO decision now, whatever they decide is good for me

azurebrd:
@nuin Thanks for trying the center center thing :)

PR #935, opened 2021-12-08
scrum-618 fix key warning

FB-ping:
add key for News component to fix the warning


azurebrd (commented):
That's a weird key value, it's the URL, right ?  Using a count didn't work ?

FB-ping:
yes, it is the url of that news, as long as unique.  counter value may also be good.

azurebrd:
@FB-ping Did you try what I suggested on the other pull request ?  Github code doesn't handle backticks well, I'll tag you on the comment I mentioned it in #scrum-blueteam.   I think doing it with a key that refers to the type of element, and the count will communicate more about what the div is for.
`<div key={``post ${count}``} className={style.postContainer}>`

FB-ping:
will try tonight.

PR #937, opened 2021-12-08
Scrum 618 news fix key

FB-ping:
use counter as part of key


azurebrd (approved):
Cool, thanks @FB-ping

azurebrd:
live now

PR #938, opened 2021-12-09
SCRUM-957 Added max-width and padding to an extra container property …

azurebrd:
…in css, section colors matching mod colors per mod to css for a section border, solid color border with radius and padding to section.

Not sure yet how to move resources to be a vertical section on the right spanning about + news on an x-large screen.

### DIFF
--- a/apps/main-app/src/containers/modLanding/About.js
+++ b/apps/main-app/src/containers/modLanding/About.js
@@ -3,10 +3,10 @@ import style from './style.scss';
 import PropTypes from 'prop-types';
 import LinkToMOD from './LinkToMOD';
 
-const About = ({htmlContent, modShortName, linkToMod}) => {
+const About = ({htmlContent, modShortName, linkToMod, sectionStyle}) => {
   return (
-    <div className='container'>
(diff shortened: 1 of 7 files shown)


azurebrd:
New code should take care of SCRUM-957

nuin (approved):

nuin:
From a code perspective, it looks good. And from a design perspective I think that's what they want, so all in all a perfect PR.

azurebrd:
Thanks @nuin !

azurebrd:
live now

azurebrd:
@valearna fyi: removed `<div className='col-md-8'>` so the news text would line up with the about text in wordpressNews.js  if you really think it should go back to that we can discuss

nuin:
This is the main reason I don't like borders

image

The pages gets unbalanced, "boxes" have different sizes, and will have for all MODs, and you lose standardization, allowing for unbalance on what is presented for each mode. But management loves it, so we are happy.

thanks for working on this @azurebrd

nuin:
Another example is the `privacy-warranty-licensing` page

PR #942, opened 2021-12-10
SCRUM-633: Point icons on homepage Members section to MOD landing pages

gildossantos:
It's perhaps too early for this change, but I would please like some feedback on the approach.

### DIFF
--- a/apps/main-app/src/containers/homepage/MembersSection.js
+++ b/apps/main-app/src/containers/homepage/MembersSection.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import {Link} from 'react-router-dom';
 
 import style from './style.scss';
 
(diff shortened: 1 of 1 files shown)


azurebrd (approved):
That looks good to me, although I haven't tried running the code to check it works.

gildossantos:
Thanks Juancarlos! Are you always up this early? :)
It does work when I run it locally, but I'm honestly not sure if that would work in a real situation.

azurebrd:
I see you're importing 
import {Link} from 'react-router-dom';

I'm not sure you need that if you're doing
          <a className={style.member} href='/members/rgd'>

I would think if you're using Link, you might want 
          <Link className={style.member} to='/members/rgd'>

But I'm not sure.  @nuin any thoughts ?

azurebrd:
haha, no, I was so tired I fell asleep around 8 pm and woke up around midnight.  I'm hopefully going back to bed soon to be up for the stand up :)

gildossantos:
ah - I was trying a lot of stuff and the import of Link is a remnant of that. I'll try removing it and the re-test. Thanks for catching that!

azurebrd:
Not at all !  On the Lit curation I'm using Link.  I'm honestly not sure when it should be Link vs regular <a>  It might have to with using history with Router ?  I imagine to keep things consistent <a> is good, unless we have a specific reason to switch.

FB-ping:
only when you render something, you will need to import "Link" from 'react-router-dom'; here this file only serve as "constant", no render, so no need to import, will need in the file which render those links.

azurebrd:
Ah, thanks @FB-ping

gildossantos:
Thanks @FB-ping and @azurebrd! I think I sorta understand that explanation - I'm sure it'll become clearer with more study and practice.
BTW, I removed the Link import and it still works on my local machine (expected I guess).

gildossantos:
Just noticed some small issues (e.g., using MGD not MGI). Will close this draft PR and fix all issues before submitting a new PR that resolves the issues. Again, thanks!

nuin:
@gildossantos Be aware that there was a discussion about MGI/MGD some time ago on the general channel on Slack

gildossantos:
Thanks @nuin - I found that thread. I'll leave it alone :)

PR #943, opened 2021-12-10
Adding MOD abbr back to browser tabs

gildossantos:


gildossantos:
woah! must've branched off the wrong one.

PR #945, opened 2021-12-10
Use "WB" as the modShortName for "WormBase" (currently short name is same as fullname)

gildossantos:
DO NOT MERGE UNTIL MORE DISCUSSION AT BLUE TEAM DAILY SCRUM

### DIFF
--- a/apps/main-app/src/containers/modLanding/content.js
+++ b/apps/main-app/src/containers/modLanding/content.js
@@ -12,7 +12,7 @@ export const MODContent = {
       'WormBase is a founding member of the Alliance of Genome Resources Project.' +
       '</p>',
     link: 'https://wormbase.org',
-    modShortName: 'WormBase',
+    modShortName: 'WB',
     modFullName: 'WormBase',
     bannerStyle: style.banner__WB,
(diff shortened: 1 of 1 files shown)


nuin:
@gildossantos I won't be at the scrum today.

azurebrd:
I think @valearna did that on purpose, but I don't remember why.  It made sense when we were refactoring.  I guess if he's not around, I'd be fine with it if Ranjana's okay with it (as a WB rep and SME for mod landing pages)

gildossantos:
hmm - I guess we use full-length "WormBase" on the Members section of the homepage too. Maybe it was intentional.
But, in that case, "FB" should also be "FlyBase" then.
And I'm sorry if I'm reopening issues that have already been discussed and resolved.

azurebrd:
Not at all :)  I'm good with it ending up either way. (All abbreviations or just some)

gildossantos:
Closing, as I think we need more info on what the MODs want, then we need to take a fresh look across the board at places that we use abbr vs fullname and make sure we implement it according to specs.

PR #946, opened 2021-12-10
SCRUM-633

gildossantos:
I think we might as well get this change into the blue branch (assuming all is ok here), but no rush.


azurebrd (approved):
Sounds good :)

azurebrd:
live now

PR #948, opened 2021-12-10
start of tests

ianlongden:
Start of tests for worm base landing page.


azurebrd (approved):
Cool.  I probably don't completely get what's happening, but it makes sense.  Some of those tests will have to change for some layout stuff Ceri wanted, but like you said, if you get it in first, I'll have to update the tests :)

azurebrd:
live

PR #950, opened 2021-12-11
More news has less size now. Or should it be fewer (pixels)?

nuin:
Title


azurebrd:
Hey @nuin I've been making changes to this with Shannon for the past 3 hours or so, and we've been changing a lot of stuff, including this, so I hope you don't mind if we don't create a merge conflict with my approving your changes.  Could you wait until I finish, and then see if you like that ?

PR #951, opened 2021-12-11
SCRUM-957 more changes to css and layout.  Importing fonts.  About li…

azurebrd:
…nk to MOD is a button now with same color stylting as the titlebar.  Tests probably fail, will fix later.  resource section paragraphs have less margin-bottom.  h4 news have more padding-top.  more news link lowercased and more vertical padding.  section text uses Open Sans with more padding and default font size instead of reduced.

### DIFF
--- a/apps/main-app/src/containers/modLanding/About.js
+++ b/apps/main-app/src/containers/modLanding/About.js
@@ -4,14 +4,14 @@ import PropTypes from 'prop-types';
 import LinkToMOD from './LinkToMOD';
 import HeadMetaTags from '../../components/headMetaTags';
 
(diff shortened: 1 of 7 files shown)


azurebrd:
Thanks for the feedback @nuin .  I think this takes care of the "more news" pretty well.  The sections have better padding and line height and a bunch of stuff (I can't recall which things changed in this PR versus the previous one).  I've been staring at this too long, but I think it looks pretty good.

nuin (approved):

azurebrd:
Thanks @nuin ! (live now)

PR #952, opened 2021-12-13
SCRUM-976: Use FlyBase and WormBase in place of FB and WB

gildossantos:
Also moving browser tab tags to Main.js from About.js.


azurebrd (approved):

azurebrd:
live

azurebrd:
@gildossantos The URL still says wb/fb  and I think that's fine.  I don't know whether it matters or not, or whether we prefer this or not.

PR #953, opened 2021-12-13
SCRUM-977: Use "MGD" consistently rather than "MGI".

gildossantos:
To prevent future confusion, used "MGD" for "MGI" even in internal variable names.
The MGD "About" text was also update, as per Carol's text.


azurebrd (approved):

azurebrd:
live

azurebrd:
@gildossantos The style.scss still has a titleBar__MGI value, which makes a couple of colours not come through for MGD

PR #955, opened 2021-12-13
fix switched rodent hex logos

gildossantos:
The wordpress site used as the source for the MOD hex logos has the logos switched for the rodents (i.e., the "mgd.png" is actually the RGD rat, and the "rgd.png" is actually the MGD mouse. If you run this branch locally, and compare it to the current public PROD alliance homepage, you'll see that the logos used in this branch do indeed match what's on the public homepage. I've added a comment in the code to explain the png naming mix up. Ultimately, the wordpress site should be fixed, and/or perhaps we should use assets in this repo for the hex images, rather than external links?


azurebrd (approved):
Sounds good.

azurebrd:
live

PR #958, opened 2021-12-15
SCRUM-633

gildossantos:
Use Link instead of href so that links from homepage Members section go to the top of the MOD landing pages (important when browsing on phone or narrow browser window).


nuin (approved):
TBH, I don't see any differences trying both branches, but the code is fine.

azurebrd:
live

PR #963, opened 2021-12-16
footer initial template and data for WormBase

ianlongden:
Adding early as it is my first venture into react etc. WormBase done so just need to add other members, but you can look at this and decide on any style changes etc that may be wanted. Just giving people advance warning.

### DIFF
--- a/apps/main-app/src/containers/modLanding/FooterAlt.js
+++ b/apps/main-app/src/containers/modLanding/FooterAlt.js
@@ -1,0 +1,41 @@
+import style from './style.scss';
+import PropTypes from "prop-types";
+
+const FooterAlt = ({links, footerStyle, modShortName}) => {
+  let generateHtml = (links, modShortName) => {
+    let htmlString = '<div class="container-fluid">' +
+                     '<div class="row">' +
+                     '   <div class="col-med-2 col-md-2 col-lg-2 col-xl-2">' +
+                     `     ${modShortName}` +
+                     '  </div>';
+    let counter = 0;
+    for (let key in links){
+        htmlString += '<div class="col-med-2 col-md-2 col-lg-2 col-xl-2">';
+        htmlString += `      <a href="${links[key]}">${key}</a>`;
+        htmlString += '</div>';
+        counter += 1;
+        // Add a blank cell under mod name to keep spaces looking okay.
+        if(counter == 5){
+            htmlString += '<div class="col-med-2 col-md-2 col-lg-2 col-xl-2"></div>';
+            counter = 0;
+        }
+    }
+    return htmlString;
+  }
+
+  return (
+      <div className={`${footerStyle} ${style.titleBar}`}>
+        <div className={`container ${style.titleBarContainer}`}>
+          <div className={style.titleBarText} dangerouslySetInnerHTML={{__html: generateHtml(links, modShortName)}}/>
+        </div>
+      </div>
+  );
+}
+
+FooterAlt.propTypes = {
+  links: PropTypes.object.isRequired,
+  footerStyle: PropTypes.string.isRequired,
+  modShortName: PropTypes.string.isRequired
+}
+
+export default FooterAlt;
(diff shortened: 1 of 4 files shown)


azurebrd (dismissed):
Generalizing the labels to links makes sense.  I think @nuin did the resources as hardcoded html to leave it open for a wordPress import in the future, so I'll let him comment on that.

azurebrd (commented):
@ianlongden Actually, I'm wondering whether it'd be better if footer_obj was an array of objects, then the order would be explicit, but also we could add other parameters in the future, like font color, whether they open in the same tab or another tab, if some should be bigger font for some reason, and other stuff like that.

ianlongden:
> @ianlongden Actually, I'm wondering whether it'd be better if footer_obj was an array of objects, then the order would be explicit, but also we could add other parameters in the future, like font color, whether they open in the same tab or another tab, if some should be bigger font for some reason, and other stuff like that.


yep. Good call an array would keep the order. I will change that.

nuin:
I would change the CSS before merging, the font colours really don't stand out

image


but it's your call @azurebrd

azurebrd:
@ianlongden I agree with @nuin , I hadn't looked at what it actually looked like.  I imagine we'd want to share the same background color + font color styling that comes from titleBar__<mod>.  It'll probably need a tweak to apply to the <a> elements.  

Separately, at some point we might end up splitting up .titleBar__<mod> into something like backgroundColor__<mod> and fontColor__<mod>  because I was tweaking adding a box-shadow to the about button, but didn't want to apply it to the title bar by using the same .titleBar__<mod> for the button styling.  This might end up mattering more as we use the same colours for other elements.

azurebrd:
@nuin Do you have an opinion about my suggestion of it being an array of objects instead of key-value pairs ?   I see that @ianlongden has made it an array of arrays, which works, but it forces tracking of index, instead of key.  I thought the content.js could be something like
```footer: [
      { 'label': 'Nomenclature', 
        'link': 'https://wormbase.org/about/userguide/nomenclature#gi9m5c264k8be0afldh71j3--10'},
      {'label': 'Tools', 
        'link':  'https://wormbase.org/tools#0--10'],
    ...]```
In case we later need to add 'target' or 'fontColor' or something else.
This totally works as-is though.

azurebrd (dismissed):

nuin:
@azurebrd I think right now both approaches are valid, but I agree with your point and suggestion, it gives more flexibility for future changes.

azurebrd:
Thanks @nuin

azurebrd:
@ianlongden If you wanted to try to update the styling for the links, I'd probably do something like in the resources, where there's a div with style resourceDiv  and the style.scss has an entry `.resourceDiv p`  which applies the styling to only the <p> elements that are within the (only-one) elements with a resourceDiv class.  So you'd probably add a  footerDiv style to a div in the footer container, then make a `.footerDiv a` in the style.scss that gives it the colour properties you'd want.  Since you'd potentially want different font colours for the different mods, you'd probably make it a footerDiv__<mod> instead of just footerDiv, and then have separate sections for each mod with a <space>a in the style.scss  (I might have made this sound more complicated than it is, I'm sure there's other ways of doing this)

nuin:
@ianlongden @azurebrd There is also some refactoring that can be done on `style.css` as new classes for the footer can be added to avoid collision between title bar and footer. And these classes can be together in the CSS

azurebrd (approved):
Thanks for taking this on @ianlongden and going along with so much backseat coding.  It's your last day, so let's get this merged and out of your mind.  I agree with @nuin about refactoring the style.scss although I'm not sure what's best, so I hope Paulo can take that on after this is merged.

azurebrd:
live

azurebrd:
@nuin If you take on the style refactor, could you also add something to make the <a> text color match the expected font color ?  If not, I can do it when you're done

nuin:
@azurebrd I will take a look, we need to override the global `a` style, I will take a look

azurebrd:
@nuin Ah, gotcha, I didn't realize that.  Thanks.

ianlongden:
I am happyish with what is there now. Need to change text color and add actual icons for twitter/standford etc but it should be okay as a MVP. Please feel free to change what ever you want while I am away,.

PR #964, opened 2021-12-17
scrum-995 add static link for flybase, zfin, goc'

FB-ping:
sorry, keep forget NOT to merge into main branch. static news link for flybase, zfin, goc here.


nuin (approved):
Codewise is fine, I would change a couple of things on sizing and other aspects. I will merge and take a look later.

azurebrd:
live.

Thanks @FB-ping and @nuin .  Probably best if we're consistent in content.js about whether fields without a value are explicitly ''/0 (like wordpressBaseURL and fetchNewsCount in flybase, or skipped (most other mods for those fields).  I don't have a preference either way, and it's fine if we're not consistent.

azurebrd:
Also agree that the link styling could look better / more consistent some other way, thanks @nuin

PR #967, opened 2021-12-18
Some changes

nuin:
increased visit button
changed links in footer to white
changes "news links" for something more explanative

### DIFF
--- a/apps/main-app/src/containers/modLanding/FooterAlt.js
+++ b/apps/main-app/src/containers/modLanding/FooterAlt.js
@@ -16,7 +16,7 @@ const FooterAlt = ({links, note, footerStyle, modShortName}) => {
                    '<div class="row">';
     for (let i=0; i< links.length; i++){
         htmlString += '<div class="col-med-2 col-md-2 col-lg-2 col-xl-2">';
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
Code looks good.  I'm not sure about about making the Footer <a> by style color white, since MGI has a light coloured background that will need a darker colour.  If possible, using the titlebar styles would be best (if you're still planning to refactor the style.scss to avoid duplication of the same values for different classes that basically mean the same thing).  I haven't tried to change it though, when you mentioned the global style, I realized it might not be that straightforward.

azurebrd:
Tests are failing, don't know if you'd want to disable part of them while still making changes, and add them back later.

azurebrd:
@nuin Adding this to the style.scss fixes the <a> font color applying it just to the <a> within the Footer.
```
.modFooter__WB a {
  color: #85A6BA;
}

.modFooter__MGD a {
  color: #dfeffe;
}
```
(and so on for other MODs).  Possibly to clean up the css we should have some variables for the color and background-color, and refer to those in the classes.

nuin:
I will change and pull again

PR #968, opened 2021-12-18
Some changes

nuin:
Couple of commits, changes are explained on the commit message

### DIFF
--- a/apps/main-app/src/containers/modLanding/FooterAlt.js
+++ b/apps/main-app/src/containers/modLanding/FooterAlt.js
@@ -16,7 +16,7 @@ const FooterAlt = ({links, note, footerStyle, modShortName}) => {
                    '<div class="row">';
     for (let i=0; i< links.length; i++){
         htmlString += '<div class="col-med-2 col-md-2 col-lg-2 col-xl-2">';
-        htmlString += `      <a href="${links[i][1]}" style="color: white">${links[i][0]}</a>`;
+        htmlString += `      <a href="${links[i][1]}">${links[i][0]}</a>`;
         htmlString += '</div>';
     }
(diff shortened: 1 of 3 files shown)


azurebrd (approved):
That's great @nuin

azurebrd:
@nuin Did you do something to fix the tests ?  I didn't get why they weren't passing before.  Wondering if something needed updating (I didn't see a change to the files for it), or if github was just messing up before, and now it isn't.

azurebrd:
live.

Looks good.

I think the News section when it has just a link, would look better if it kept the "News" heading to look like the ones that an API, but the ticket doesn't say that clearly. https://agr-jira.atlassian.net/browse/SCRUM-626  I'll ask Ceri on the ticket to confirm.

The Footer doesn't say we should use the mod hexagonal icon, but I think that would look more consistent with the Alliance footer.  I don't think it should have the mod name, but I'm less sure about that until trying it / seeing it.  The ticket doesn't explicitly say not to, so I think that'd be good.

Footer could also use icons (like Twitter, YouTube).  

I'll make this comments on the relevant tickets.  @nuin Do you think these are good ideas ?  I don't know if you're planning on making more changes (also, are you still thinking of refactoring the style.scss for the background-color and color ?)  If so, that's great, if not, let me know.

nuin:
I will work on it this evening, we can coordinate to test the display after

azurebrd:
@nuin cool cool.  If at any point you'd like me to take over any part of it, let me know.  No rush.

nuin:
I think the icon looks better on the footer than on the title bar to be honest, so we can get rid of the MOD name from it.

azurebrd:
@nuin Sounds good.

PR #971, opened 2021-12-22
fixing project name and version on package.json

nuin:


azurebrd (approved):
Thanks for catching that @nuin

PR #983, opened 2022-01-10
Last chance for mocks

ianlongden:


azurebrd (approved):
That's great @ianlongden .
It would probably be good to have a meeting (others welcome) about how testing works, to get on the same page about it.
I see   jest.mock('./../../../../src/lib/fetchData');  and apps/main-app/src/lib/__mocks__/fetchData.js   does jest know to check for /__mocks__/ versions when something gets 'jest.mock'ed ?
What does useLayoutEffect do ?
Is 'data-testid' something that is required for testing, that is, it can't work off of a regular 'id' that could be used for other things in the code ?
Everything about the testing it() part makes sense.

azurebrd:
live

azurebrd:
Also did an npm install before restarting, to have the new packages so npm test passes.

PR #998, opened 2022-01-13
New footer tests for each member

ianlongden:


azurebrd (commented):
@ianlongden For what it's worth, the index variable in FooterAlt.js is basically the same as count, but it's not pre-incremented so the data-testid _count value would be off by one.  I'm still looking at the rest of the tests, although I'll probably only look carefully at one Mod, not all

nuin (approved):
Is there a way to have to have on footer test file for all? Just curious.

azurebrd (changes requested):
@ianlongden I think it'd be better to change the <mod>_footer.test.js to use i instead of i+1 in the screen.getByTestId, and _0 for the First item ;  to match a change of FooterAlt.js to avoid having a separate count variable, and using index in lines 26 and 28 instead of count.   I think it would make the code more concise and more directly match the indices of values getting expect-tested (screen.getByTestId vs toHave Attribute).  I hope that make sense.  If you think that won't work, that's okay, and I can approve this.

azurebrd:
@nuin I guess it would require having an array of mods + key-values for each expected thing being tested for ?  I imagine for footers that's possible, but for things that are API-based it would be more complicated (they return different format json, and some don't have a field others have).  This new testing way that Ian's refactored is much better than the old one that had a manual copy of the values in the content.js that are being tested for.

azurebrd (approved):
Great, thanks @ianlongden !

azurebrd:
live

PR #1002, opened 2022-01-14
SCRUM-1026 googleapis has a start and end date, the start date is inc…

azurebrd:
…lusive, but the end date is exclusive, so it's one more day than we want.  SGD uses pacific timezone, so set that when creating the date object.  Changed date range format to match SGD webpage.


nuin (approved):

azurebrd:
Thanks @nuin .  live.

PR #1005, opened 2022-01-18
Basic whole page tests. Containers done separately so this is just ma…

ianlongden:
…king sure they play nicely together


azurebrd:
live

PR #1006, opened 2022-01-18
Check static link for news

ianlongden:


azurebrd (approved):
Cool, thanks @ianlongden   So we just need to test one of the MODs ? (to know whether to only test one of the meetings links)

azurebrd:
live

PR #1010, opened 2022-01-19
test for the about section in mod landing pages SCRUM-1040

dougli1sqrd:
for https://agr-jira.atlassian.net/browse/SCRUM-1040


ianlongden:
So I like the way you pass short test data instead of reading the content.js that makes for shorter more precise tests. The only thing I am worried about is doing the test in another way to the other tests. We also have to import more libraries.
I think I also prefer the way other tests use data-testid's and the tests for children are much easier to understand as it is just expect(element1).toContainElement(element2) than the way they are done here.
Would you mind using these forms of tests?
data-testid is more specific and indicates to the developer which items are important and if changed need the test changed etc.
Please feel free to persuade me otherwise if there is anything that the new way is advantageous over the current way.

dougli1sqrd:
Yeah I can use the test ID impl with the other testing library. I was just trying that since I had heard that that was the more "modern" way so just wanted to play with it. 

I'm glad you like the using the short test data instead of using content.js. I fear using content.js ends up, in part, testing the content itself, so our tests will fail if the content changes.

nuin:
I don't think we are merging this one, correct? The link for the test is `google.com`. But if you guys want I can merge this, up to you.

nuin:
BTW, just @ianlongden can approve this according to @dougli1sqrd, so whatever you decide is fine.

azurebrd:
@ianlongden @dougli1sqrd I thought we wanted the tests to fail if the content changes.  I'm fine with it either way.
Like @nuin I also find it odd that it's testing for www.google.com and the tests pass.

dougli1sqrd:
I mean if the content changes that means one of us went in and chnaged the content. If we test the content, doesn't that mean we're just duplicating the content, and we have to keep two files in sync?

azurebrd:
I don't honestly get the point of testing, so just modeling stuff after what Ian's written and asking questions to see if it becomes clearer.

dougli1sqrd:
@ianlongden updated.

@azurebrd ha interesting. How come? It's definitely true UI testing is pretty confusing and doesn't conform to more standard backend style testing procedures. Do you think we just don't need to test the UI?

azurebrd:
@dougli1sqrd Oh, I just meant, we have a component that renders some element inside some other element.  So the test renders the component, searches for both elements, and tests that one is inside the other.  It's the same code generating the component, so I'm not sure what the test accomplishes.  In that example, if we change the component, we have to change the test.  In loading stuff from the constants.js we change the content of the constants.js so then we also change the test.  It seems really circular, and in general I don't get what it would catch, other than someone mistakenly deleting a bunch of lines, the page somehow still rendering, and no one catching it during code review.  I just haven't done any testing before we started scrum.  (Well, I wrote some unit tests for some Alliance codebase [maybe ferret ?], but the task was "increase the coverage" not "understand this in a meaningful way")

dougli1sqrd:
Yeah it's true. In this simple of component and test I tend to agree with you. I guess if we had special interactions or logic built into the components that would be a more ripe area to test. But this is pretty circular, especially testing the content like that.

I wonder if we should we change our testing DOD/guidelines specifically in a UI context?

PR #1016, opened 2022-01-20
SCRUM-1086 Added Search section that only links to Alliance search fo…

azurebrd:
…r specific species.  Cleaned up some commented out stuff

### DIFF
--- a/apps/main-app/src/containers/modLanding/Main.js
+++ b/apps/main-app/src/containers/modLanding/Main.js
@@ -8,6 +8,7 @@ import Title from "./Title";
 import News from "./News";
 import Meetings from "./Meetings";
 import Resources from "./Resources";
+import Search from "./Search";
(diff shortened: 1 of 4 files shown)


ianlongden (approved):
I do not see any tests for this. (you knew it was coming :-) ). If this is because it is experimental and is just while people look at it to see if it is acceptable then it is okay. But I will check later on these added. :-)

azurebrd:
@ianlongden for sure :)  I think they won't quite like how it looks, and we'll have to change the formatting, which might mean having the code diverge from how the resource container handles the content.js data.  Thanks !

azurebrd:
@ianlongden I guess UIPR/BuildAndTest has to pass before it can merge on master, while it didn't on blue_team_code ?

azurebrd:
live.

PR #1018, opened 2022-01-20
SCRUM-1086 updated test div data-testid to refer to the right column …

azurebrd:
…instead of the resource column


azurebrd:
@ianlongden Proper tests are coming.  For a bit it looked like Search didn't work (it doesn't work for species), but Ceri found a workaround, and Paul likes it, so this should be good.

azurebrd:
Tests are in.  Unless people want different links or styling, this should be done.

nuin (approved):

azurebrd:
Thanks @nuin .  live.

PR #1025, opened 2022-01-25
Release 4.2.0 into master

oblodgett:


azurebrd (changes requested):
Hi @oblodgett 
I think @valearna 's still on leave, so I don't know when he'd get around to reviewing this.
I think the removal of lines from apps/main-app/src/constants.js would get rid of the mod landing pages links from the menubar and the footer, but those are also using outdated routes, so I'm not sure.  I don't know how 4.2.0 merging with master after the blue_team_code merge should resolve.  If you think this is correct, then I'm fine with not having changes.
Separately, Blue team has been talking about removing the package-lock.json from the repo, but let us know if you think we shouldn't.

oblodgett:
So there are a couple of bug fixes that went into 4.2.0 and also the `package-lock.json` is a generated file so it should be excluded.

azurebrd:
Cool, thanks @oblodgett

nuin:
@oblodgett Do you want us to remove `package-lock.json` from the branch/versioning?

azurebrd:
@nuin I think Blue team wants to remove it, and @oblodgett is okay with us doing that.

oblodgett:
@azurebrd @nuin @valearna this is a standard PR to bring in all the change from 4.2 into master.

azurebrd (approved):

azurebrd:
Thanks @oblodgett .  I don't recall requesting changes, and I couldn't tell what changes it was talking about, so I've approved the requested changes, and it's approved the PR.  I don't know what the lift failure is about.

PR #1028, opened 2022-01-28
SCRUM-1029 Added RGD News and Meetings API.  Reusing NewsFlybase for …

azurebrd:
…News, since format is the same.  Could try to reuse if knew how to pass a news/meetings string to parseFlybasePosts to get appropriate api response.


nuin (approved):

azurebrd:
Thanks @nuin .  live.

PR #1032, opened 2022-01-31
SCRUM-1138 branch has typo, it should be this ticket.  Make a new Vis…

azurebrd:
…it Mod section to put below resources, take the Visit Mod link out of the About and add it to the h2 header of the new Visit Mod section


azurebrd:
Thanks @nuin

PR #1050, opened 2022-04-06
SCRUM-1064 Pathway widget queries GO-CAMs by UniProtKB for human genes

dustine32:


valearna (approved):
I'm not familiar with the pathway widget, but the changes look good to me

PR #1054, opened 2022-05-11
SCRUM-1219 fix

christabone:
Updated spelling of sonicParanoid to fix missing data.


christabone:
I think the adjustment actually needs to be made in the loader. We've modified algorithm names in the past and this should be added to the list.

christabone:
I'm going to merge this temporarily and then revert it once the data is fixed in the backend. That will allow folks to at least test the site for now even if the spelling is incorrect.

PR #1059, opened 2022-08-05
Scrum 1564 Adding Xenbase Logo

pjhale:
This pull adds an updated header image with the Xenbase logo attached.  In addition adds the Xenbase logo to the members section above the footer.  The link for this goes to Xenbase, unlike the other mods but should be trivial to change in the future once the Xenbase Alliance mod landing page is created.  Note... the Xenbase individual hex logo does appear to be slightly different graphically although with help from @motenko it is now the correct size and lines up with the other icons.


azurebrd (approved):
Cool, thanks @pjhale

azurebrd (approved):
Thanks @pjhale !

PR #1060, opened 2022-08-12
Added Xenbase to header & footer

pjhale:
Xenbase added to header and footer via the constants file.

No visible issues with css on both homepage and gene pages.  Both the header and footer are correctly showing the "link out" symbol for the href to xenbase home.  The ticket didn't specify if Xenbase had a longer form label (ex Mouse Genome Database vs MGD) so they are both the same right now.

### DIFF
--- a/apps/main-app/src/constants.js
+++ b/apps/main-app/src/constants.js
@@ -121,6 +121,11 @@ export const NAV_MENU = [
         route: '/members/wormbase'
       },
       {
+        label: 'Xenbase',
+        shortLabel: 'Xenbase',
+        route: 'https://www.xenbase.org/'
+      },
+      {
         label: 'Zebrafish Information Network',
         shortLabel: 'ZFIN',
         route: '/members/zfin'


valearna:
thanks @pjhale! Is the dev site online? https://dev2994.alliancegenome.org/ returns "bad gateway"

valearna (approved):

azurebrd (approved):
I don't think we have any use cases of Xenbase wanting XB displayed anywhere.  A new category "prefix" or "curie prefix" is planned for the database, but if Xenbase doesn't need it on the UI, there's no need to create it in the constants.js

@pjhale @valearna It'd probably eventually be good to dockerize launching the UI on the devserver, so that instead of one developer running it in their own screen, someone could deploy it, and someone else could stop it and redeploy it.  Then we could always leave it running.  For now Paul could keep it running in a screen for others to look at, but we could also talk at standup about what Ceri would want (and then we'd document it so that we don't forget it's running like I did)

pjhale:
I currently don't have the dev server running... I've got two different branches that I need Ceri to sign off on.  I think talking a bit at stand up to see whats the best way to use this server would be good.

azurebrd:
@pjhale makes sense

PR #1074, opened 2022-09-14
Adding case to add chr prefix to Laevis JBrowse Links

pjhale:
Added a catch that checks for Laevis taxon and the presence of a chr  prefix.  If this doesn't exit will add.  We can remove this when the  chromosome for laevis matches the JBrowse chromosome.


scottcain (approved):
As much as I hate this type of hack, I think it's what's needed for now. We can reevaluate how chromosome names are handled at a later date.

christabone (approved):

PR #1075, opened 2022-09-14
the increadingly unwieldy test for modifying location chromosome strings

scottcain:

PR #1076, opened 2022-09-15
Fix for laevis scaffold issue

pjhale:
Fixes the JBrowse links for Laevis scaffolds being given the chr prefix. 
 Also fixes the display string so these entries don't say 
"ChrScaffold1234".


christabone (approved):

pjhale:
@scottcain Can you take a look at this?  This fix didn't address the lack of data from apollo for these genes but I think you were addressing that in another ticket.

scottcain:
Yeah, I have this PR: https://github.com/alliance-genome/agr_ui/pull/1075 which I think should fix the apollo thing

scottcain (approved):
👍

PR #1078, opened 2022-09-29
Revert algorithms for Xenbase release

christabone:


cmpich (approved):
Looks good, though, I thought, we had taken those two out of the mix.

christabone:
@cmpich Xenbase is using some older DIOPT algorithm data that we're loading up for this release.

PR #1102, opened 2023-03-03
made both type and page.name lowercase when looking up URLs

adamgibs:


apps/main-app/src/components/dataTable/getResourceUrl.js
```diff
@@ -20,7 +20,7 @@ export const getResourceUrl = (curie, type) => {
   } else {
     [resource] = resourceDescriptors
         .filter(resource => resource.db_prefix === prefix)[0].pages
-        .filter(page => page.name === type.toLowerCase())[0].url
+        .filter(page => page.name.toLowerCase() === type.toLowerCase())[0].url
```
oblodgett, line 23: page.name is already lowercase? What does this fix?

adamgibs: No all of them. The one that was failing for zfin was "Fish"

oblodgett: Ahh ok... this is wrong in the resource Descriptor file:
https://github.com/alliance-genome/agr_schemas/blob/master/resourceDescriptors.yaml#L183

adamgibs: So should this still go in? It want hurt anything if this is later fixed in the agr_schemas

PR #1116, opened 2023-04-07
removed slash and index.html. Maybe this will fix this 🤷‍♂️

scottcain:
hopefully will address https://agr-jira.atlassian.net/browse/KANBAN-318

### DIFF
--- a/libs/app-shell/src/lib/sitemap.js
+++ b/libs/app-shell/src/lib/sitemap.js
@@ -24,7 +24,7 @@ const sitemap = [
       },
       {
         label: 'JBrowse',
-        route: '/jbrowse/index.html?data=data%2FHomo%20sapiens',
+        route: '/jbrowse?data=data%2FHomo%20sapiens',
       }
     ],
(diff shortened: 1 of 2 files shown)


libs/app-shell/src/lib/sitemap.js
```diff
@@ -24,7 +24,7 @@ const sitemap = [
       },
       {
         label: 'JBrowse',
-        route: '/jbrowse/index.html?data=data%2FHomo%20sapiens',
```
oblodgett, line 27: @scottcain this goes in the sitemap xml file for google to use so you might not want to change this if the direct hit in a browser still works.

scottcain: This is actually reverting a change I made last week when I didn't understand the nature of the problem.

oblodgett: oh ok

oblodgett (approved):

PR #1118, opened 2023-04-14
SCRUM-2783

adamgibs:
This works locally. We'll see if it does on stage

### DIFF
--- a/apps/main-app/src/routes.js
+++ b/apps/main-app/src/routes.js
@@ -54,6 +54,15 @@ export default (
         }}
       />
 
+      <Route
+        exact
+        path='/jbrowse'
+        render={() => {
+          window.location.href = '/jbrowse/?data=data%2FHomo%20sapiens';
+          return null;
+        }}
+      />
+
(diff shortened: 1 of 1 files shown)


apps/main-app/src/routes.js
```diff
@@ -54,6 +54,15 @@ export default (
...
 
+      <Route
+        exact
+        path='/jbrowse'
+        render={() => {
+          window.location.href = '/jbrowse/?data=data%2FHomo%20sapiens';
+          return null;
+        }}
+      />
+
```
oblodgett, line 65: Ahh... didn't know there was a "route" for blue genes. This makes sense... @scottcain

oblodgett (approved):

oblodgett:
@adamgibs can this work for ALL jbrowse links?

adamgibs:
@oblodgett you mean like the one in the Alleles and Variants section of the gene page? I think so since it's treated as an external link. That said, there very well may be some issues with this solution for other Jbrowse links that I didn't catch. Are there any other site links that match "/jbrowse" exactly that shouldn't link to human?

PR #1120, opened 2023-04-14
SCRUM-2783 added home to browser history in jbrowse route

adamgibs:


oblodgett (approved):
I am guessing this is so the back button works correctly?

adamgibs:
@oblodgett Yep, it seems to work on my local machine this  way.

PR #1121, opened 2023-04-14
SCRUM-2783 made jbrowse an external link in site menu

adamgibs:


libs/app-shell/src/lib/sitemap.js
```diff
@@ -97,10 +99,14 @@ const sitemap = [
       {
         label: 'Alliance User Community',
         route: 'https://groups.google.com/a/alliancegenome.org/g/community',
+        external: true,
+
```
oblodgett, line 103: @adamgibs these are not served in the app these are generated in the sitemap.xml file no need to add this here.

apps/main-app/src/constants.js
```diff
@@ -93,6 +93,7 @@ export const NAV_MENU = [
       {
         label: 'JBrowse',
         route: '/jbrowse/?data=data%2FHomo%20sapiens',
+        external: true,
```
oblodgett, line 96: Can this be added to both bluegenes and api links?

adamgibs: Yeah, sure.

oblodgett: I'm guessing that external here means external to react not necessarily external to the site.

adamgibs: Yeah, specifically using ExternalLink component

apps/main-app/src/routes.js
```diff
@@ -52,14 +52,6 @@ export default (
         }}
```
oblodgett, line 52: Im guessing you can remove these other routes also? Given external=true?

adamgibs: On my local machine the api works fine this way, but I get a page not found with Alliance Mine.

oblodgett: Thats fine... just trying to be consistent with the links.

oblodgett: There is proxies also setup in the UI... so it might be good to check them also for alliance mine... that is why the API is working its being proxyed

adamgibs: Do you mean the dev server proxies in the webpack.config?

oblodgett (approved):

PR #1127, opened 2023-05-24
SCRUM-2727 enable filtering on reference column for disease association table on gene page

cmpich:


oblodgett (changes requested):
Needs to go in the release branch.

PR #1129, opened 2023-05-24
SCRUM-2394: fix import check GeneticSex

abecerra:
There is another problem in the table in my local:

`Uncaught TypeError: Cannot read properties of undefined (reading 'split')
    at getResourceUrl (getResourceUrl.js:4:30)
    at getSingleReferenceCurieAndUrl (utils.js:75:29)
    at SingleReferenceCellCuration (singleReferenceCellCuration.js:8:57)`

### DIFF
--- a/apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
+++ b/apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
@@ -19,6 +19,7 @@ import RelatedNotes from './RelatedNotes';
 import EvidenceCodesCellCuration from './evidenceCodesCellCuration';
 import AnnotationSource from './AnnotationSource';
 import AnnotationType from './AnnotationType';
+import GeneticSex from './GeneticSex';
 
 function renderLink(entity) {
   // Console.log(entity);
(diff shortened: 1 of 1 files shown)


apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
```diff
@@ -84,7 +85,7 @@ function AnnotatedEntitiesPopupCuration(props) {
                     <td><ExperimentalConditionCellCuration conditions={entity.conditionRelations}/></td>
                     <td><ExperimentalConditionCellCuration conditions={entity.conditionModifiers}/></td>
                     <td><StrainBackground strainBackground={entity.sgdStrainBackground}/></td>
-                    <td><GeneticSex geneticSex={entity.geneticSex}/></td>
+                    <td>{entity.geneticSex && GeneticSex(geneticSex=entity.geneticSex)}</td>
```
oblodgett, line 88: You can just add a entity.geneticSex? and it will not fail but then you need to check it in the component.

oblodgett (changes requested):

apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
oblodgett: Plus the error is not here... its in the component. `startsWith` doesn't exist.

abecerra: added entity.geneticSex?

oblodgett: The error is not coming from the AnnotatedEntitiesPopupCuration.js file: its coming from the GeneticSex component.

abecerra:
I commented all the annotation details, leaving <td>text</td> for the entries. Now, the problem in the console is:
```
utils.js:66 Uncaught TypeError: Cannot read properties of undefined (reading 'startsWith')
    at indexWithPrefix (utils.js:66:18)
    at getRefString (utils.js:41:7)
    at getSingleReferenceCurieAndUrl (utils.js:74:17)
    at utils.js:80:40
    at Array.map (<anonymous>)
    at getMultipleReferencesCuriesAndUrls (utils.js:80:21)
    at Object.ReferencesCellCuration [as formatter] (referencesCellCuration.js:17:63)
```

abecerra:
Closing because the import fix was already merged yesterday

PR #1131, opened 2023-05-29
Add sequence panel

scottcain:


scottcain:
I haven't tested this PR myself yet (I have to read up on how to create a test environment). I guessed at what would be needed looking at existing code.

apps/main-app/src/constants.js
```diff
@@ -248,6 +248,9 @@ export const SPECIES = [
     apolloName: 'human',
     apolloTrack: '/All%20Genes/',
     jBrowseName: 'Homo sapiens',
+    jBrowsenclistbaseurl: 'https://s3.amazonaws.com/agrjbrowse/docker/5.4.0/human/',
+    jBrowseurltemplate: 'tracks/All_Genes/{refseq}/trackData.jsonz',
+    jBrowsefastaurl: 'https://s3.amazonaws.com/agrjbrowse/fasta/GCF_000001405.40_GRCh38.p14_genomic.fna.gz',
```
oblodgett, line 253: One thing that we might want to do is put cloudfront in front of this S3 bucket... it will speed up the delivery of these files and it costs less.

scottcain: That's (I assume) a reasonable idea. I don't know anything about cloudfront, so I'll defer to you.

scottcain: In any event, it should impede testing.

PR #1134, opened 2023-05-30
SCRUM-2394 asserted genes and notes updates

adamgibs:
-- asserted genes display symbol instead of curie
   -- main row gene is filtered out of asserted genes
   -- notes are prefixed with either note or summary depending on type

### DIFF
--- a/apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
+++ b/apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
@@ -39,7 +39,7 @@ function renderLink(entity) {
 }
 
 function AnnotatedEntitiesPopupCuration(props) {
-  const {children, entities} = props;
+  const {children, entities, mainRowCurie} = props;
 
   if (!entities || !entities.length) {
     return null;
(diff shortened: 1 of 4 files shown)


apps/main-app/src/components/dataTable/AnnotatedEntitiesPopupCuration.js
```diff
@@ -81,7 +81,7 @@ function AnnotatedEntitiesPopupCuration(props) {
                   <tr key={entity.subject.curie}>
                     <td>{renderLink(entity)}</td>
                     <td><TypeCellCuration subject={entity.subject}/></td>
-                    <td><AssertedGenes assertedGenes={entity.assertedGenes}/></td>
+                    <td><AssertedGenes assertedGenes={entity.assertedGenes?.filter(gene => gene.curie !== mainRowCurie)}/></td>
```
oblodgett, line 84: @adamgibs can this logic be moved into the component?

oblodgett: Send both assertedGenes and gene to the component and let it figure it out.

oblodgett (changes requested):
@adamgibs can this logic be moved into the component?

adamgibs:
@oblodgett logic moved into component

oblodgett (approved):

PR #1140, opened 2023-06-12
SCRUM-2723 removed subtype hack code

adamgibs:


oblodgett:
@adamgibs I am guessing you have tried this against the stage API?

adamgibs:
@oblodgett Yep, and the example page I mentioned this morning is working now. Looks like all subtypes are coming through at least with the examples I've tested

oblodgett (approved):

PR #1152, opened 2023-07-12
SCRUM-2703: sort by number of predictions methods matched

abecerra:


markquintontulloch (approved):
This appears fine to me in that it brings the behaviour of the paralogy table in line with that of the orthology table, but it doesn't address Chris' request to have a secondary sort based on bestScore and bestScoreRev (unless I'm missing something).  I would be inclined to have a sort function that can be called by both tables that deals with this.

abecerra:
> This appears fine to me in that it brings the behaviour of the paralogy table in line with that of the orthology table, but it doesn't address Chris' request to have a secondary sort based on bestScore and bestScoreRev (unless I'm missing something). I would be inclined to have a sort function that can be called by both tables that deals with this.

I added a sorting function with the bestScore and bestScoreRev. It leaves the entries with (Yes,Yes) at the top, but it might need fixing

PR #1153, opened 2023-07-12
SCRUM-2823 add disease qualifier column to disease annoation table on…

cmpich:
… gene page

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,6 +90,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.9.6",
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -90,6 +90,7 @@
   },
   "devDependencies": {
     "@babel/core": "7.9.6",
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
```
oblodgett, line 93: @adamgibs I thought we figured this issue out?

adamgibs: Not that I'm aware of. This is the only solution that worked for Christian and I.

oblodgett (approved):

PR #1158, opened 2023-08-03
Gene page Sequence panel

scottcain:


adamgibs (approved):
@scottcain Looks good to me. One thing to think about in the future is potentially refactoring the sequencePanelWrapper to be functional component instead of a class-based one. In general, we try to write functional components since that's the way modern React components are usually written (and with hooks for state). That said, there still are class-based components sitting around in the code and as long as it works, that's the most important thing.

scottcain:
@adamgibs that's funny--when I was adding the wrapper, I just looked at another file "at random" (`genomeFeatureWrapper.js`) and mimicked what it was doing. Turns out that needs updating too. So it goes :-)

scottcain:
Forgot to link in this issue: https://agr-jira.atlassian.net/browse/SCRUM-2879

PR #1166, opened 2023-08-04
SCRUM-2372

adamgibs:


apps/main-app/src/components/dataTable/utils.js
```diff
@@ -79,6 +79,72 @@ export const getMultipleReferencesCuriesAndUrls = (references) => {
...
+  })
+}
+
+export const buildSourceUrl = (dataProvider) => {
+  if(!dataProvider) return;
+  let url;
+
+  //default to mod homepage if no crossReference is provided
+  if(!dataProvider.crossReference){
+    url = dataProvider.sourceOrganization?.homepageResourceDescriptorPage?.urlTemplate?.replace("[%s]", "");
```
oblodgett, line 139: Does this take care of the home page? I'm not sure you would need to do a string replace here?

oblodgett (approved):

apps/main-app/src/components/dataTable/utils.js
adamgibs: It's needed for SGD. Their homepage URL from the API is `http://www.yeastgenome.org/[%s]`

PR #1178, opened 2023-08-25
Kanban 366

scottcain:
This PR does two things:
updates the URLs for JBrowse data to the 6.0.0 release (should be abstracted out in the future)
adds a fix for yeast and frog chromosome names to prefix 'chr' when needed.


scottcain:
Also, I should note that I wasn't able to run locally since the yeast data was updated to complete the fix for KANBAN-366, but given how it was working before that data was updated, I'm confident this PR will work the same.

PR #1186, opened 2023-09-11
Added code for deploing a ALB with the UI code

oblodgett:


mluypaert:
@oblodgett You may want to generalise the `StageALBStack` class to a `publicSiteALBStack` and then extend that into `StagePublicSiteALBStack` and `ProdPublicSiteALBStack` for respectively stage and prod envs (using a few parameters). Otherwise you'd need to copy-paste this class for the Prod deployment, while most of it should be identical.

oblodgett:
@mluypaert yes... definitely want to do this. I was going to have Chris put in another ticket... but I also wanted testing for this ALB to go forward, to make sure I didn't miss anything. I had to have commits across multiple repos, gocd, nginx, so just want to make sure it all works at this point. Another issue was I couldn't figure out a rule in the ALB to redirect if someone it hit directly. As in stage-alb.* -> stage.*

oblodgett:
Plus at the moment not sure how to have two "configs" in the same cdk, I don't think its hard but again was just trying to get something working last night. I think you can have mulitple cdk "apps" and then you just specify the one you want to run.

mluypaert:
> @mluypaert yes... definitely want to do this. I was going to have Chris put in another ticket... but I also wanted testing for this ALB to go forward, to make sure I didn't miss anything. I had to have commits across multiple repos, gocd, nginx, so just want to make sure it all works at this point.
> Another issue was I couldn't figure out a rule in the ALB to redirect if someone it hit directly. As in stage-alb.* -> stage.*

I think you can add a rule to the ALB that uses a condition where if host-header equals the `stage-alb.` URL directly that it redirects to `stage.alliancegenome.org`. The tricky part is going to be finding what property to use for that stage-alb URL, as it uses pseudo-random names for the actually resources CDK deploys (in other words you can not and should not hard-code the stage-alb URL).

oblodgett:
I tested this last night and the name that always comes through is "stage-alb" so anything coming through the ALB or directly to the ALB has the hostHeader of "stage-alb" so thats not an options.

PR #1215, opened 2023-11-07
Remove gene name from column header on interaction table, KANBAN-386

motenko:


oblodgett (approved):

chris-grove:
@motenko 
Could we discuss this change? This is a reversion of a specifically requested feature of the Interactions Working Group

PR #1216, opened 2023-11-10
SCRUM-2393: Allele disease annotations (UI)

abecerra:


cmpich (approved):
Looks good.

oblodgett:
@cmpich this breaks the other tables as @abecerra has mentioned.

@abecerra I think the main table looks good but I think you might want to not reuse / change the components that the table uses but if you do need to change them then just create new ones with Curation on the end of the file name. This was what was done on the other tables.

abecerra:
> @cmpich this breaks the other tables as @abecerra has mentioned.
> 
> @abecerra I think the main table looks good but I think you might want to not reuse / change the components that the table uses but if you do need to change them then just create new ones with Curation on the end of the file name. This was what was done on the other tables.

No @oblodgett , I restored the component that is used in phenotypes for the gene page and pushed my changes in a different file. Maybe in the future we could review and refactor them.

PR #1233, opened 2023-12-20
Update deps

adamgibs:
Update UI to use CRA


oblodgett:
@adamgibs i doubt these are your errors but i think we should try to clean them up before merging.

abecerra (commented):
After deleting node_modules, npm build throws:

internal/modules/cjs/loader.js:818
  throw err;
  ^

Error: Cannot find module 'webpack'

adamgibs:
@oblodgett Yeah, the create react app eslint settings are set up to catch different things so now things in the older code is being caught. I can fix some of the issues now. However, I think the anchor tag errors should be their own ticket since that's more of a refactor.

adamgibs:
@abecerra Try this:

- delete node_modules
- delete package-lock.json
- run npm install --legacy-peer-deps
- run npm run build
- run npm start

abecerra:
> @abecerra Try this:
> 
> * delete node_modules
> * delete package-lock.json
> * run npm install --legacy-peer-deps
> * run npm run build
> * run make uirunstage

I forgot to delete package-lock.json. Gene and allele pages look good!

abecerra (approved):

abecerra:
Maybe it would be  worth preserving the .editorconfig file?

adamgibs:
@abecerra That's a good point. I don't know if the `.editorconfig` file would work the same in a CRA app. It's worth looking into.

oblodgett (approved):

PR #1242, opened 2024-01-11
SCRUM-3604: render Add Impl Genes and Str Background only in Gene page

abecerra:


oblodgett (commented):
Does this help with the code reuse vs having two components?

abecerra:
> Does this help with the code reuse vs having two components?

Yes, the same component is called in the Gene and Allele page. 

But, I don't know about the AnnotatedEntitiesPopup component (the old) and the other pages that are using it.

adamgibs (approved):
This works for now I think. In the future, if we need to use more and more conditional logic like this, it may make sense to refactor this component into two or more components that share core logic.

PR #1247, opened 2024-01-29
SCRUM-3644: Reorder logic for filters

abecerra:


adamgibs (approved):

src/components/dataTable/DataTable.js
```diff
@@ -96,7 +96,7 @@ const DataTable = ({
   }
 
   // the initial fetch is happening, show spinner, don't render the table yet
-  if (isLoading || isFetching) {
```
oblodgett, line 99: I believe you want the isFetching because when they hit the second page this needs to turn on.

oblodgett (changes requested):
Has this been tested on all Data Tables on the site?

abecerra:
@oblodgett: without the `isFetching`, the pagination+filter issue is solved. I am not sure why 
I tested in the gene and disease pages (with the disease ribbon)

oblodgett:
We need to have the isFetching so somewhere there we need to fix it... this need to not be a hack fix

oblodgett (approved):

PR #1249, opened 2024-02-07
bumping seqpanel urls to 7.0.0 (should really parameterize!)

scottcain:


src/constants.js
```diff
@@ -252,7 +252,7 @@ export const SPECIES = [
     apolloName: 'human',
     apolloTrack: '/All%20Genes/',
     jBrowseName: 'Homo sapiens',
-    jBrowsenclistbaseurl: 'https://s3.amazonaws.com/agrjbrowse/docker/6.0.0/human/',
+    jBrowsenclistbaseurl: 'https://s3.amazonaws.com/agrjbrowse/docker/7.0.0/human/',
```
oblodgett, line 255: @scott could you set this value before the `export const SPECIES = [` then you would only have to change it in one place?

scottcain: Probably. I could
```
export const RELEASE = '7.0.0';
```
and then later on
```
jBrowsenclistbaseurl: `https://s3.amazonaws.com/agrjbrowse/docker/${RELEASE}/human/`,
```
but I'm not sure of the syntax. Are backticks interpolating?

adamgibs: Yeah, that syntax will work. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

adamgibs: Also, you don't need the export if you're using the variable in the same file.

oblodgett (approved):

PR #1257, opened 2024-02-16
KANBAN-490

adamgibs:


bin/build_robots.js
```diff
@@ -0,0 +1,28 @@
...
+Allow: /
+Disallow: /search
+Disallow: /alliancemine
+Disallow: /jbrowse
+Sitemap: https://www.alliancegenome.org/sitemap.xml
+Sitemap: https://www.alliancegenome.org/api/sitemap.xml`;
+
+const OTHER_CONTENT = `# https://www.robotstxt.org/robotstxt.html
+User-agent: *
+Disallow:
```
oblodgett, line 16: I think this should be 
```
User-agent: *
Disallow: /
```
I know thats not what is on stage but reading this site: https://searchfacts.com/robots-txt-allow-disallow-all/

package.json
```diff
@@ -66,8 +66,9 @@
     "build": "react-scripts build",
     "test": "react-scripts test",
     "eject": "react-scripts eject",
-    "prebuild": "npm run resource-descriptors",
+    "prebuild": "npm run resource-descriptors && npm run robots",
```
oblodgett, line 69: Can you not have two items on the line? `npm run resource-descriptors robots`

bin/build_robots.js
adamgibs: Ah, sure. I can do that.

package.json
adamgibs: Hmm, I'm not sure. I can try it though.

oblodgett: Plus this can be in the post build doesn't have to be in the prebuild.

adamgibs: The double ampersand appears to be necessary unfortunately .

PR #1262, opened 2024-02-27
updating the generic-sequence-panal package to deal with SARS-CoV-2

scottcain:
(gives a clean error now at least). Addresses https://agr-jira.atlassian.net/browse/KANBAN-520

Would it have been better to have this in 7.0.0? yes, but it snuck up on me. The PR represents a bandaid to give a clean error message with SARS-CoV-2 sequences until I can re-architect the sequence panel to work more generically.


oblodgett:
@scott is this going into the release or not... we are trying to do that today?

scottcain:
Yeah, I suppose that would be better but probably not a showstopper either.

oblodgett (approved):

PR #1270, opened 2024-02-28
fair chance I got the proxy settings for jbrowse2 right

scottcain:
adding proxy settings for jbrowse2 so I can start testing code with it.


scottcain:
Doesn't have to be merged right away if you want to make sure the 7.0.0 release has settled.

oblodgett (approved):

PR #1282, opened 2024-03-07
adding a few proxy lines to allow fetching RGD and XenBase data

scottcain:


cdk/amplify-stage-stack.ts
```diff
@@ -18,8 +18,12 @@ export class AmplifyStageStack extends cdk.Stack {
...
       { source: '/jbrowse2',                           target: 'https://stage.alliancegenome.org/jbrowse2/',                                          status: amplify.RedirectStatus.PERMANENT_REDIRECT },
-      { source: '/jbrowse2/',                          target: 'https://stage.dgaayxgqoarxf.amplifyapp.com/',                                       status: amplify.RedirectStatus.REWRITE },
-      { source: '/jbrowse2/<*>',                       target: 'https://stage.dgaayxgqoarxf.amplifyapp.com/<*>',                                    status: amplify.RedirectStatus.REWRITE }, 
+      { source: '/jbrowse2/',                          target: 'https://stage.dgaayxgqoarxf.amplifyapp.com/',                                         status: amplify.RedirectStatus.REWRITE },
+      { source: '/jbrowse2/<*>',                       target: 'https://stage.dgaayxgqoarxf.amplifyapp.com/<*>',                                      status: amplify.RedirectStatus.REWRITE }, 
+      { source: '/jbrowsedata/XenBaseXTJBrowse/<*>',   target: 'https://jbrowse.xenbase.org/XenJBrowse/data/xt9_1/<*>',                               status: amplify.RedirectStatus.REWRITE },
```
oblodgett, line 23: @scottcain you will need to have 3 entries for each of these... without the tailing `/` which is a redirect, then the one with the tailing `/` and then the wildcard... otherwise the browser does funny things.

oblodgett (changes requested):

cdk/amplify-stage-stack.ts
scottcain: I'm not so sure that's true--this is what each of those configs looks like in the browser editor at the JBrowse amplify:
```
    {
        "source": "/jbrowsedata/XenBaseXTJBrowse/<*>",
        "target": "https://jbrowse.xenbase.org/XenJBrowse/data/xt9_1/<*>",
        "status": "200",
        "condition": null
    },
 ```
which works--I suspect it works because it's not a "web" fetch but a "data" fetch, if that makes sense.  Or should the `status` be 200 in the cdk config?

oblodgett: Oh, is this always a call from javascript to /jbrowse2/<something>? Then yes this should be fine. The issue is if you goto `/jbrowsedata/XenBaseXTJBrowse` or `/jbrowsedata/XenBaseXTJBrowse/` in a browser it wont work.

scottcain: Yes this is just for JS requests for specific files. There is no reason anyone would ever make a request for /jbrowsedata/RGDJBrowse/index.html and I think they should return a 404

scottcain: Actually, if you try to navigate to https://stage.dgaayxgqoarxf.amplifyapp.com/jbrowsedata/RGDJBrowse/mRatBN7.2/ you'll get a 403 and if you try https://stage.dgaayxgqoarxf.amplifyapp.com/jbrowsedata/RGDJBrowse/mRatBN7.2/index.html you get a 404. I think both are fine and expected--those urls will never exist anywhere on the site (not in the main site nor in JBrowse 2), so the only way to navigate to there is for a person/spider to see that a URL like https://stage.dgaayxgqoarxf.amplifyapp.com/jbrowsedata/RGDJBrowse/mRatBN7.2/Animal%20Diseases%20Related%20Genes.gff3.gz exists, and then try trimming off the file name to see what they get. I think 403 or 404 are perfectly good results in that case.

PR #1293, opened 2024-04-02
adding region highlighting for variants linked out to JBrowse

scottcain:
Reasonably sure this works but need to test in stage environment to make sure.


scottcain:
To be clear--the reason I hedged a bit in my comments above is that to really test that it is working, I need the stage site so that links to JBrowse 2 will work (since there is no "http://localhost:3000/jbrowse2" to link to. There is no reason to fear my breaking something :-)

adamgibs (approved):

PR #1295, opened 2024-04-03
improving the buffer around genes when highlighting a variant

scottcain:
Adds some space on the sides when linking to a highlighted variant (this buffering is done with other JBrowse links but wasn't done here, so fixing that),


oblodgett (approved):

scottcain:
that was fast! :-)

PR #1296, opened 2024-04-08
SCRUM-3743 UI work to display disease page associated alleles from persistent store

VarunReddy1111:
Also fixed errors in displaying gene table


adamgibs (approved):

cmpich (changes requested):
I can see changes for the alleles table which should not get into stage yet. Only the changes for the gene table yet.

PR #1298, opened 2024-04-09
Fixed gene table in disease page according to API fields

VarunReddy1111:


src/components/dataTable/utils.js
```diff
@@ -17,6 +17,19 @@ export const getDistinctFieldValue = (response, field) => {
       value && value.trim()
     ));
 };
+export const getIdentifier = (subject) => {
+  if(!subject) return;
+  return subject.curie?subject.curie:(subject.modEntityId?subject.modEntityId:subject.modInternalId);
```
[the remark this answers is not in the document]
VarunReddy1111: sure

src/components/dataTable/utils.js
```diff
@@ -17,6 +17,19 @@ export const getDistinctFieldValue = (response, field) => {
...
+}
+
+export const getGeneObject = (subject) => {
+  if(!subject) return;
+  const gene = {
+    curie: getIdentifier(subject),
```
cmpich, line 28: I am not sure about the usefulness of this method which only returns a slimmed down version of a gene object, just curie and symbol.
I would rather replace 'gene.curie' occurrences with
getIdentifier(gene)

VarunReddy1111: Makes sense, will remove the method

cmpich (approved):

PR #1300, opened 2024-04-17
SCRUM-3919 Disabling the download button for data tables with more than 90000 rows

VarunReddy1111:


src/components/dataTable/DataTable.js
```diff
@@ -61,6 +63,13 @@ const DataTable = ({
     return filters;
   };
 
+  let disabled = false;
+
+  const setDisabled = (value) => {
+    disabled = value;
+    return true;
+  };
```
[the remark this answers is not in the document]
VarunReddy1111: Tried using useState, but leading to infinite renders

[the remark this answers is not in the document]
adamgibs: I _think_ you could just initialize the variable like this:

`  let disabled = paginationObj.options?.totalSize > ROW_THRESHOLD;`

It would reinitialize each render and then you would not need the setter function or the logic inline with the JSX.

src/constants.js
```diff
@@ -431,3 +431,4 @@ export const GA_EVENT_ACTION = {
   SET_PAGE_SIZE: 'Set page size'
 };
 
+export const ROW_THRESHOLD = 90000;
```
oblodgett, line 434: This should be called for what it is: DISEASE_GENES_MAX_ROWS or something better like that.

VarunReddy1111: This is applicable to all the data tables, would it be better to have more generic name like DATA_TABLE_MAX_ROWS?

oblodgett: This should not be for all datatables it should only before this table?

oblodgett: Ahh ok... so we at least need to alert Chris that this is going to be a global change and not just for this table, and see if he is good with it. Then I would call it DOWNLOAD_BUTTON_THRESHOLD or something like that.

PR #1305, opened 2024-04-19
adding proxy entries for JB2 for test and production

scottcain:
Of course, there's no really good way to test this until it actually goes on production, but I'm reasonably sure I got this right.


oblodgett (commented):
These need to go into test so that they can go out before the release?

scottcain:
If you're asking if they need to be in the 7.1.0 release, then yes, we have to do whatever needs to be done to make that happen. I don't know the flow of this code. Should I make an additional PR into test as well or instead?

oblodgett (approved):

PR #1315, opened 2024-05-03
SCRUM-3841

adamgibs:


oblodgett (approved):
Does this need to go with API changes also?

adamgibs:
@oblodgett Yeah, that's a good point. I can hold off until Christian merges in his code.

PR #1345, opened 2024-07-25
KANBAN-584 pavi integration

mluypaert:
Enabling intergration of constants and related functions into the PAVI UI (to enable reused of jbrowse configs) through the updates in this PR.


oblodgett (commented):
Typically the .js is omitted when doing includes for JS.

mluypaert:
> Typically the .js is omitted when doing includes for JS.

Ya, while I am aware of that, doing so means I cannot import this modules through it's raw github https URL, because the relative paths don't resolve without their extensions. Including the extension does work though in all cases (both for local paths as through URL imports).

mluypaert:
And I need to import through URL as these files are not local to the PAVI code, nor deployed and installable through npm (and probably should not be either).

adamgibs (approved):

motenko (approved):

PR #1347, opened 2024-07-29
JBrowse AGR release constant removal

mluypaert:
@scottcain This should solve the issue with the hard-coded `RELEASE` constant getting outdated on every new alliance release. It does assume new jbrowse data is released with every new AGR release though (but that behavior can be overridden through the `REACT_APP_JBROWSE_AGR_RELEASE` env variable.


scottcain (approved):
I haven't tested this on my machine, but it feels like a very reasonable approach. Thanks for taking care of this.

PR #1348, opened 2024-07-30
Main branch mergeback

mluypaert:


oblodgett:
@mluypaert why is this getting generated? We already do this as part of the release.

oblodgett (approved):

mluypaert:
> @mluypaert why is this getting generated? We already do this as part of the release.

@oblodgett I created this mergeback PR manually after this code was merged in production as I thought having the most recent jBrowse data available on stage would be beneficial for testing the then upcoming (7.3.0) release on the stage environment (before release to production).

mluypaert:
> Yes but when we click the "update branch" button this automatically happens. So I don't think there is a need for this mergeback PR as the code is already merged back.

Now, yes. But note that I opened this PR 2 weeks ago, at which point testing on the stage environment was still ongoing without having the latest jBrowse data available there (due to the bug fixed here). So getting this merged then could have been beneficial, and the "update branch" button approach would only be available when creating the PR to create the production release (2 weeks later).

PR #1357, opened 2024-08-20
KANBAN-598 - Add more resource links to AD portal

motenko:


pjhale (approved):
Looks good... I dont know if the first link is supposed to be "Alzheimer's" instead of "Alzheimer" but the results it pulls back look fine.

PR #1383, opened 2024-12-19
Scain kanban 643 646

scottcain:
Should address both https://agr-jira.atlassian.net/browse/KANBAN-643 and https://agr-jira.atlassian.net/browse/KANBAN-646

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@testing-library/jest-dom": "^5.16.1",
     "@testing-library/react": "^11.2.5",
     "@testing-library/user-event": "^13.5.0",
+    "abortcontroller-polyfill": "^1.7.8",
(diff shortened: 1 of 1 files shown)


scottcain:
The build failure does not appear to be related to the version change for the sequence panel

mluypaert:
> The build failure does not appear to be related to the version change for the sequence panel

While it didn't seem that way, the error was a result of the sequence-panel upgrade, because some of the AGR code relied on a library that was not defined as a direct dependency of the web code, but got installed as a subdependency of the older @jbrowse/core library. I've fixed that in [873bfc6](https://github.com/alliance-genome/agr_ui/pull/1383/commits/873bfc6373e52a90031d566f25a3e1a3a98cbebb).

mluypaert (approved):

PR #1391, opened 2025-01-14
KANBAN-591

adamgibs:


src/components/expression/expressionComparisonRibbon.js
```diff
@@ -58,7 +58,7 @@ const ExpressionComparisonRibbon = ({
   let updatedSummary;
   if (summary.data) {
     const taxonIdYeast = 'NCBITaxon:559292';
-    const categories = summary.data.categories.filter(category => !(
+    const categories = summary.data?.categories?.filter(category => !(
```
sethrcamp, line 61: A few thoughts here:

1) Since this is wrapped in `if (summary.data)`, then you shouldn't need the first "?" (i.e. `summary.data.categories?..."

2) "categories" can be undefined if `data.categories` is undefined. Just curious if this is intentional.

3) This is a pretty minor optimization, but both "selectedOrthologs" and "geneTaxon" are not dependant on the categories array, meaning `selectedOrthologs.length === 0 && geneTaxon === taxonIdYeast` will be the same for every category. Rewriting it this way will reduce the number of times the comparison will be made to 1 (or O(1) to be fancy):
```
let updatedSummary;
if (summary.data) {
  const taxonIdYeast = 'NCBITaxon:559292';
  let categories = summary.data.categories;
  if (categories && selectedOrthologs.length === 0 && geneTaxon === taxonIdYeast) {
    categories = categories.filter(category => !category.id.startsWith('UBERON:');
  }
  ...
}
```

4) Lastly, `updatedSummary = summary.data;` will be a pointer reference, meaning `updatedSummary.categories = categories;` will change the value for both "summary" and "updatedSummary". I don't recall if react query supports this, but either way, if this isn't intentional, I would change to this instead:
```
updatedSummary = {
  ...summary.data,
  categories
};
```

src/containers/genePage/alleleTable.js
```diff
@@ -23,12 +22,12 @@ const AlleleTable = ({ isLoadingGene, gene, geneId}) => {
...
   const tableProps = useDataTableQuery(`/api/gene/${geneId}/alleles`);
   const {
-    resolvedData,
+    data: resolvedData,
     isLoading,
   } = tableProps;
 
   const data = useMemo(() => {
-    return resolvedData ? resolvedData.results?.map(allele => ({
+    return resolvedData ? resolvedData?.map(allele => ({
```
sethrcamp, line 30: Here's a little shorthand if you like avoiding ternaries:
```
return (resolvedData || []).map(allele => ... );
```

sethrcamp (commented):
Left a few thoughts, but overall looks pretty straight forward (although I am sure it looks much more straightforward than it was to actually upgrade everything).

adamgibs:
@sethrcamp Pushed out code to address your comments. Thanks!

motenko (approved):
looks good

PR #1409, opened 2025-02-24
SCRUM-3995 phenotype annotations on gene and allele page

cmpich:


src/setupProxy.js
```diff
@@ -5,6 +5,7 @@ module.exports = function(app) {
 				'/api',
 				createProxyMiddleware({
 						target:	 process.env.API_URL || 'http://localhost:8080',
+						timeout:60000,
```
oblodgett, line 8: Is there a reason for this change? The UI code handles timeouts in a different place.

adamgibs (approved):

PR #1410, opened 2025-02-24
Address security vulnerability

mluypaert:
Adresses dependabot security vulnerability: https://github.com/alliance-genome/agr_ui/security/dependabot/104


mluypaert:
Nvm I misread the dependabot information, 4.6.2 is still vulnerable to problem, and 5.* is not compatible with the currently used reactstrap dependency version. Closing this ticket.

PR #1411, opened 2025-02-25
Enabled dependabot vulnerability reporting for subdependencies

mluypaert:
The lack of the `package-lock.json` file in the github repository makes dependabot unable to report on known (public) vulnerabilities in subdependencies, which is not desired behavior as we need to be able to detect and fix vulnerabilities, even when they occur in subdependencies. This PR reintroduces that file to (re)enable dependabot vulnerability alerts for all dependancies, so we can detect and fix them as required.


oblodgett (changes requested):
This needs to go into stage as we then role it forward.

mluypaert:
> This needs to go into test as we then role it forward.

Now address (I updated the base branch to be test).

oblodgett (approved):

PR #1414, opened 2025-02-25
Kanban 383

Joe-Wdz:
Add vert scroll to nav section


oblodgett (changes requested):
Something isn't right with the pull request... I suggest you rebase onto the end of stage, this looks like you are going to revert all the code from last release. Looks like you took the branch from `test` which would mean you need to merge back into `test`.

PR #1423, opened 2025-03-05
KANBAN-674 Fix spelling of Alzheimer

motenko:


src/containers/alzheimersPage/ResourcesSection.js
```diff
@@ -5,10 +5,10 @@ import ExternalLink from '../../components/ExternalLink';
 
 const ResourcesSection = () => {
   const resourceLinks = [
```
sethrcamp, line 7: This is a really small optimization, but `resourceLinks` should be moved outside of the component. i.e.:
```
const resourceLinks = [...];

const ResourcesSection = () => {...};
```
The reasoning is that even though it never changes, react will instantiate the variable on each render of the `ResourcesSection` component.

Not a big deal here, but if you were to have something that called a chain of helper functions for example, it could add up to a noticeable performance impact, specifically if the component is used multiple times on a page.

sethrcamp: Also, further down within the render, the <div> should have a key attribute. 
```
<div key={"something-unique-here"}>
```
I'd have to look it up, but there is a `useId` hook meant exactly for this purpose, but I am not sure which version of React added it.

Even still, you will have to do something like this:
```
const id = useId();
...
resourceLinks.map((resourceLink, index) => (
<div key={`${id}-${index}`}>
...
```
That way each div gets a unique key on the page.

motenko: Thanks @sethrcamp - I'll give it a try.

motenko: I've addressed these concerns with new commits - thanks for the suggestions.

sethrcamp:
This looks good to me!

sethrcamp (approved):
This looks good to me!

jimflybase (approved):

PR #1426, opened 2025-03-10
Kanban 517

Joe-Wdz:
KANBAN-517, ? for help in a different location.


motenko (changes requested):
As part of the refactor, could you also delete the two files you replaced ... GoControlsHelp and ExpressionControlsHelp?

Joe-Wdz:
> As part of the refactor, could you also delete the two files you replaced ... GoControlsHelp and ExpressionControlsHelp?

Thanks Howie, I should have deleted it after refactoring.

motenko (approved):

PR #1431, opened 2025-03-12
Kanban 686

sethrcamp:
Replaces sequence viewer with under construction message.

I created a flag for clarity, but it is likely something we will never actually toggle, but remove entirely once Apollo is back up and running


sethrcamp:
Keeping this a draft for the moment so I can rebase and get rid of the unnecessary commit chain

motenko (approved):

PR #1432, opened 2025-03-13
Kanban-529

jimflybase:


motenko (changes requested):
Jim, I saw your comment about the social media icons. I'm not sure - but can you check to see if SocialMedia is used anywhere else - probably not.  If this is the only place, would you please delete the file? If that makes sense.  Thanks, Howie

sethrcamp (approved):

motenko (approved):

PR #1448, opened 2025-03-26
SCRUM-4009 update disease summary links

adamgibs:


src/containers/diseasePage/index.js
```diff
@@ -41,10 +41,10 @@ const DiseasePage = ({diseaseId}) => {
 
   const title = data.doTerm.name || data.doTerm.curie;
 
-//  let definitions = [data.definition];
-//  if (data.doTerm.definitionUrls && data.doTerms.definitionUrls.length > 0) {
-//    definitions.push(data.doTerm.definitionUrls[0]);
-//   }
+ let description;
+ if (data.doTerm?.definitionUrls?.length > 0) {
+   description = data.doTerm.definition + " " + data.doTerm.definitionUrls[0];
```
oblodgett, line 46: So the description always needs to be set to `data.doTerm.definition` and then add the first URL if there is one.

adamgibs: code has been updated

PR #1449, opened 2025-03-27
SCRUM-4009 implement alpha sort

adamgibs:


src/lib/utils.js
```diff
@@ -24,6 +24,41 @@ export function compareAlphabeticalCaseInsensitive(accessor) {
   };
 }
 
+export function alphaSort(accessor) {
```
oblodgett, line 27: can this be called smartAlphaSort

adamgibs: Name updated

oblodgett (approved):

src/containers/diseasePage/basicDiseaseInfo.js
```diff
@@ -23,7 +23,7 @@ class BasicDiseaseInfo extends Component {
     return terms &&
       <CollapsibleList>
         {terms
-          .sort(compareAlphabeticalCaseInsensitive(term => term.name))
+          .sort(smartAlphaSort(term => term.name))
           .map(term => <Link key={term.curie} to={`/disease/${term.curie}`}>{term.name}</Link>)
```
oblodgett, line 27: @markquintontulloch here is an example usage of it.

PR #1452, opened 2025-03-28
Gocamviz 1.1.2 package lock sync

mluypaert:
@tmushayahama @christabone You forgot to update or at least commit the `package-lock.json` file in [PR#1447](https://github.com/alliance-genome/agr_ui/pull/1447). Depending on how you tested these change, that may mean the installed version during testing was the old one all along. If you just forgot to commit the `package-lock.json` file updates then you can just accept and merge this PR, otherwise please test again (locally) and then accept and merge when proven to be functioning as expected.


mluypaert:
The consequence of this file not being committed would be that the test environment is still running gocamviz 1.0.1-beta2 instead of the new version.

christabone (approved):

PR #1468, opened 2025-04-11
SCRUM-5034 fix UI for orphanet

cmpich:


src/components/dataTable/utils.js
```diff
@@ -27,7 +27,7 @@ export const getIdentifier = (subject) => {
 
 export const getSingleReferenceUrl = (pubModId) => {
   let url;
-  if(pubModId.includes("PMID")){
+  if(pubModId.includes("PMID") || pubModId.includes("ORPHA")|| pubModId.includes("ORPHA")){
```
oblodgett, line 30: Does this need to be checked for twice?

markquintontulloch: @oblodgett - I believe the appropriate check is now in place

oblodgett (approved):

PR #1478, opened 2025-04-29
SCRUM-5102 remove row select from alleles table

adamgibs:


src/containers/genePage/alleleTable.js
```diff
@@ -79,27 +79,28 @@ const AlleleTable = ({ isLoadingGene, gene, geneId}) => {
...
+  //          ...alleleIdsSelectedPrev.slice(0, indexAlleleId),
+  //          ...alleleIdsSelectedPrev.slice(indexAlleleId + 1)
+  //        ];
+  //      } else {
+  //        return [...alleleIdsSelectedPrev, alleleIdRow];
+  //      }
+  //    });
+  //  },
+  //  style: { backgroundColor: '#ffffd4' },
+  //}), [alleleIdsSelected, setAlleleIdsSelected]);
```
oblodgett, line 103: @adamgibs you might be able to just leave the method uncommented and just comment out below

src/containers/genePage/alleleTable.js
```diff
@@ -313,8 +314,9 @@ const AlleleTable = ({ isLoadingGene, gene, geneId}) => {
           data={data}
           downloadUrl={`/api/gene/${geneId}/alleles/download`}
           keyField='id'
-          rowStyle={{cursor: 'pointer'}}
-          selectRow={selectRow}
+          //uncomment when sequence viewer is working again
```
oblodgett, line 317: I would put the JIRA ticket in this comment as we know we need to turn this back on in the future.

sethrcamp:
I looked into this _very_ briefly when it was mentioned yesterday, and it looked like something to do with the viewer library not having checks in place for an "invalid" id passed to it.

My only question would be: is it possible for this error to display in any of the other contexts the view is used in? If so, it may be beneficial to move the fix into the genomeFeatureWrapper.js file where there is already the flag/change that removed the viewer entirely. 

I haven't had a chance to do further testing, so it is possible that this solution is sufficient. I'll try to do some testing soon and see if it affects the other instances of the viewer

sethrcamp:
Ahh, so interestingly, the error does still happen on the allele-details page, however, because of the placement of the error boundary, the error shows where the under construction message currently is instead of replacing everything

adamgibs:
@sethrcamp Yeah, that's a good point. As you note, the issue doesn't seem to be manifesting on the alleles or variants pages that I've checked. Are there any other places that you can think of?

src/containers/genePage/alleleTable.js
oblodgett: ahh ok

oblodgett (approved):

PR #1488, opened 2025-05-14
fixing bigwig server url

scottcain:
I'm guessing Jeff would like this merged into production too for the JBrowse workshop tomorrow.


oblodgett:
@scottcain if it needs to go into all three sites... aka stage, test, and production then the PR needs to go into test then it will PR forward to production and back to stage. Sorry I should have asked what the intention of this was.

PR #1490, opened 2025-05-14
KANBAN-629 gene page jumps to different section

Joe-Wdz:
The gene page "jumps" to a different table section when changing filters on the Alleles and Variants Table.


sethrcamp:
This seems like a hacky way to do it, but I honestly can't think of a better way, and certainly not a simpler way.

I left a couple of thoughts, but it looks good to me!

sethrcamp (approved):

PR #1526, opened 2025-07-25
Scottcain fix frog sequence view

scottcain:
Needs more testing, but this might be it.


scottcain:
I didn't run prettier; I didn't see a npm command for it. Claude was mostly right, which about how much I would generally expect (the comment about `sca` for chromosome prefixes (should) only apply to frogs; I'm reasonably for other assemblies it will have no effect. In any event, I tested with a gene on a tropical scaffold and everything seemed to work correctly (in addition to checking other assemblies and several genes in leavis and tropicalis. I think this is ready to merge.

pjhale (commented):
The code changes here seem fine but its failing checks.  I'm not sure exactly why but I wouldn't want to approve without knowing.  @sethrcamp any ideas?

motenko:
@scottcain - I implemented the prettier-check, but never provided a way to 'make prettier' (in the Makefile).  At the repo root you can run "npx prettier --write ." to format your code using prettier.

scottcain:
Would somebody (@motenko ? @sethrcamp ?) mind taking a look at the status of this PR? When I run prettier now, I get no diff with git and build and serve work on my machine. Any idea why these are failing now?

adamgibs:
@scottcain You should be able to merge this in and skip the tests. Click the "Merge without waiting for requirements to be met (bypass rules)" checkbox and merge in.

scottcain:
@adamgibs I don't have that option; presumably I lack sufficient privileges (which seems perfectly reasonable to me ;-)

adamgibs:
@scottcain Ah, sure. That makes sense. I can merge it in for you.

PR #1531, opened 2025-07-31
Fixes failing test suite

sethrcamp:
Sets up additional jest config stuff.

Fixes tests failing due to upgrade changes.

Refactors redundant mod footer tests. Each test suite was identical, and the FooterAlt component has no mod-specific logic, thus the only way for all of the footer tests to not pass/fail all at the same time was if there was a problem with the ModContent object. Because that object is already checked by propTypes, having individual tests for each mod did not provide us with any additional benefit. I replaced the tests with a generic mod. This also subverts an issue with jest + vite where importing images (i.e. `import Logo from "./assets/logo.png`) does not work properly.

Updates and refactors `ModPage/test_wole.test.js`. Updates were needed due to how react router changed. The tests were written in a loop for each mod, however, the mod-specific data was not used in any of the test cases. I replaced the loop with a generic mod mock both for efficiency and to remove the dependency on content-specific test cases. I also removed the duplicate test.

Runs prettier

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,13 +1,12 @@
 module.exports = {
-  testEnvironment: 'jest-environment-jsdom', // Same name of the lib you installed
-  setupFilesAfterEnv: ['<rootDir>/test/jest.setup.ts'], // The file you created to extend jest config and "implement" the jest-dom environment in the jest globals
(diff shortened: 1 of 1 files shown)


jimflybase (approved):
Looks OK to me.

PR #1543, opened 2025-08-13
KANBAN-757 Fix Allele/Variant Table Interaction and Dependencies

christabone:
- Implemented variant selection handler from JBrowse genome feature widget
- Connected JBrowse variant clicks to the Variant table display
- Fixed data flow between the genome feature component and variant table

- Fixed genomefeatures dependency by switching from vendored version to npm package (@gmod/genomefeatures)
- Removed vendor/genomefeatures reference from Amplify build configuration
- Resolved module resolution issues that were preventing variant table interactions

- Updated "Clear Selection" button styling to match the page's blue button theme for consistency
- Added informational note to alert users when filtering is disabled while viewing selected variants
- Improved visual feedback for variant selection state

### DIFF
--- a/src/hooks/useAlleleSelection.js
+++ b/src/hooks/useAlleleSelection.js
@@ -16,7 +16,7 @@ export default function useAlleleSelection(tableProps) {
   const [isLoadingSelectedAlleles, setIsLoadingSelectedAlleles] = useState(false);
   const [selectedAllelesData, setSelectedAllelesData] = useState(null);
   const [selectedAllelesError, setSelectedAllelesError] = useState(null);
-  
+
   // Use a request counter to ignore stale responses
   const requestCounter = useRef(0);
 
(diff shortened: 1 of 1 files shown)


src/containers/genePage/alleleTable.jsx
```diff
@@ -31,10 +48,80 @@ const AlleleTable = ({ isLoadingGene, gene, geneId }) => {
...
+    }
+
+    return processedData;
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [resolvedData]);
+  }, [resolvedData, selectionOverride, selectedAllelesData]);
 
-  const [alleleIdsSelected, setAlleleIdsSelected] = useState([]);
+  // Handler for allele selection that tracks source (table vs viewer)
+  const handleAllelesSelect = React.useCallback(
```
adamgibs, line 62: I wonder if this function would be better as a hook. We usually have our code that calls the API as a hook. Maybe not something that needs to hold up the release now as long as it's working, but perhaps as an improvement for the future.

christabone: I'm working on this now, I think I can actually convert it without too much trouble.

PR #1566, opened 2025-09-18
Kanban 755 add entity counts

Joe-Wdz:


motenko (changes requested):
Do you think we should merge into test?

PR #1569, opened 2025-09-24
Kanban 803

sethrcamp:
This solves an issue from the react-bootstrap-table-next library. In a recent-ish update to one of its child dependencies, react-transition-group, it was made necessary to supply a ref to both the CSSTransition and Transition components if you are using strict mode.

The react-bootstrap-table-next is no longer being maintained, and thus does not adhere to the new standard. Because of the complex nature of the table library, it would be difficult to fix the issue through overriding components from the table library itself. Instead, a local library was made to wrap react-transition-group.

The local modules wraps both the CSSTransition and Transition components ensuring a ref is passed, even if none was provided. All libraries using react-transition-group were overridden to use the local version instead.

**NOTE:** You must run the following when pulling this code for the first time: `rm -rf node_modules && rm package-lock.json && npm install`


motenko (approved):
Look good.

PR #1731, opened 2026-04-06
SCRUM-5854: require is_a-only ancestor relationship for gene type cla…

VarunReddy1111:
…ssification

Ensure SO:0000704 ancestor must have only "is_a" relationship (not "part_of") to classify as GENE. Fixes promoter, CTCF_binding_site, and intein_encoding_region incorrectly showing as "GENE" instead of "GENOME FEATURE".

### DIFF
--- a/src/containers/genePage/index.jsx
+++ b/src/containers/genePage/index.jsx
@@ -96,7 +96,8 @@ const SECTIONS = [
 ];
 
 const GenePageCategoryLabel = ({ gene }) => {
-  const isGeneType = gene.geneType?.curie === 'SO:0000704' || !!gene.geneType?.ancestors?.['SO:0000704'];
+  const ancestorRels = gene.geneType?.ancestors?.['SO:0000704'];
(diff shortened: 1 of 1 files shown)


src/containers/genePage/index.jsx
```diff
@@ -118,7 +120,8 @@ const GenePage = () => {
   const { speciesName, taxonId, geneSymbolText, dataProviderAbbr } = extractGeneFields(gene);
 
   const ancestorRelTypes = gene.geneType?.ancestors?.['SO:0000704'];
-  const isGeneType = gene.geneType?.curie === 'SO:0000704' || ancestorRelTypes?.includes('is_a');
+  const isGeneType =
+    gene.geneType?.curie === 'SO:0000704' || (ancestorRelTypes?.length === 1 && ancestorRelTypes[0] === 'is_a');
```
oblodgett, line 124: Looks like the GENOME_FEATURE_CATEGORY needs to change?

oblodgett: aka there is no GENOME_FEATURE_CATEGORY

PR #1734, opened 2026-04-06
Fix homepage hang when Reactome script times out (KANBAN-1179)

christabone:
## Summary
- **Make Reactome script async**: Add `async` attribute to the `diagram.nocache.js` script tag in `index.html` so it no longer blocks page rendering. The homepage and all non-pathway pages now load immediately regardless of Reactome server status.
- **Add timeout to polling loop**: The `loadReactomeDiagram()` method in `pathwayWidget.jsx` now gives up after 15 seconds instead of polling forever when the Reactome library fails to load.
- **Fix while-loop condition**: The existing condition `typeof Reactome != 'undefined' && !Reactome` would skip the polling loop entirely when `Reactome` was undeclared (the exact failure case). Changed to `typeof Reactome === 'undefined' || !Reactome` so it correctly waits in both cases.
- **Show fallback message**: When the Reactome library times out, the pathway tab shows "The Reactome pathway diagram is currently unavailable" with a direct link to view the pathway on Reactome, instead of an empty diagram area.

## Test plan
- [ ] Load the homepage — should render immediately without waiting for Reactome
- [ ] Navigate to a gene page with pathway data — Reactome diagram should still load and display normally
- [ ] Simulate Reactome timeout (e.g., block `reactome.org` in DevTools Network panel) — homepage should still load; gene page pathway tab should show fallback message after ~15s with a link to Reactome
- [ ] Verify Reactome Reactions tab and GO-CAMs tab are unaffected by the change


christabone:
@claude Can you address that minor not and fix the code to pass the prettier check please?

jimflybase (approved):
Looks good to me - I trust Claude (!)

christabone:
@claude Please review again.

PR #1739, opened 2026-04-07
KANBAN-1182: Update agr_ui genomefeatures dependency

christabone:
## Summary
The genomefeatures fix for the missing human NF1 gene model has already been merged, but agr_ui was still pinned to an older GitHub commit and would not install it automatically. This update moves agr_ui to the merged genomefeatures revision and refreshes the lockfile so deployments and local installs pick up the fix consistently.

## Changes
- Updated the `genomefeatures` dependency in `package.json` to point at the merged commit `37dc2289d9193aa39823a58ee221b3c14a6d3185`.
- Refreshed `package-lock.json` so it resolves the new `genomefeatures` commit.
- Refreshed the lockfile entry for `generic-sequence-panel` to `1.8.0`, matching the existing `package.json` declaration and fixing the prior `npm ci` sync error.

## Testing
- Ran `npm install` successfully in `alliance-genome/agr_ui` to regenerate the lockfile.
- Ran `npx jest src/containers/genePage/tests/genomefeatures-import.test.js src/containers/genePage/tests/genomeFeatureWrapper.test.jsx --runInBand` successfully.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,8 +28,8 @@
         "custom-event-polyfill": "^1.0.7",
         "d3-selection": "^3.0.0",
         "events": "^3.3.0",
-        "generic-sequence-panel": "^1.7.0",
-        "genomefeatures": "github:alliance-genome/genomefeatures#a49427156f449ea11319ac2bd3069751d25ba4e5",
+        "generic-sequence-panel": "^1.8.0",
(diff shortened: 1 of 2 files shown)


christabone:
Closing — the dependency pin update here is included in #1741, which also adds the variant sequence viewer fix and regression tests.

PR #1770, opened 2026-05-19
KANBAN-1322 add disease ontology browser

jdepons:
## Summary
Adds a new disease ontology browser page at `/ontology/disease`:
- Collapsible tree view with batched term + count fetching
- Autosuggest search with full-results modal
- Term detail panel (definition, parents, synonyms, xrefs, annotation counts)
- New "Browse Ontology" entry on the disease page that deep-links into the browser at the current term

Also extends `PageNav` to support count badges and external links (used by the new disease-page entry).

Depends on **KANBAN-1322 in agr_java_software** (https://github.com/alliance-genome/agr_java_software/pull/1610) for the matching `/api/disease/batch`, `/api/disease/counts`, and `/api/disease/{id}/ancestors` endpoints. Merge that one first.

## Test plan
- [ ] `/ontology/disease` loads, top-level DOID terms render
- [ ] Clicking a term expands children; counts appear as badges
- [ ] Search box autosuggests after 2+ chars; "View all results" opens the modal
- [ ] Term detail panel shows definition, parents, synonyms, xrefs
- [ ] Disease page (e.g. `/disease/DOID:14330`) shows "Browse Ontology" in left nav and navigates to the browser at that term

Generated with Claude Code (https://claude.com/claude-code)

PR #1772, opened 2026-05-19
KANBAN-1322 add disease ontology browser

jdepons:
## Summary
Adds a new disease ontology browser page at `/ontology/disease`:
- Collapsible tree view with batched term + count fetching
- Autosuggest search with full-results modal
- Term detail panel (definition, parents, synonyms, xrefs, annotation counts)
- New "Browse Ontology" entry on the disease page that deep-links into the browser at the current term

Also extends `PageNav` to support count badges and external links. The same `PageNav` change ships in `feature/KANBAN-678` so it's a no-op in whichever PR merges second.

Depends on **KANBAN-1322 in agr_java_software** for the matching `/api/disease/batch`, `/api/disease/counts`, and `/api/disease/{id}/ancestors` endpoints. Merge that one first.

Supersedes the closed #1770 (which was based on the older pre-SCRUM-6057 stage).

## Test plan
- [ ] `/ontology/disease` loads, top-level DOID terms render
- [ ] Clicking a term expands children; counts appear as badges
- [ ] Search box autosuggests after 2+ chars; "View all results" opens the modal
- [ ] Term detail panel shows definition, parents, synonyms, xrefs
- [ ] Disease page (e.g. `/disease/DOID:14330`) shows "Browse Ontology" in left nav and navigates to the browser at that term

Generated with Claude Code (https://claude.com/claude-code)


jdepons:
Addressed all review feedback in subsequent commits — the earlier review describes the original code, not the current branch tip.

**Commit `1145ac5f` (`KANBAN-1322 address review feedback`):**
- `OntologySearchBox.jsx` Enter handler: dropped the custom `onKeyDown` that always selected `suggestions[0]`. Now relies on Autosuggest's built-in `onSuggestionSelected` (already wired up) so arrow-key navigation works correctly, with `highlightFirstSuggestion` added so Enter without arrowing still picks the first item.
- `CountBadge.jsx` unreachable code: removed the `hideZero && count === 0` branch (line above already returns for `count === 0`). Also dropped the now-unused `hideZero` prop from PropTypes and from the `OntologyTree.jsx` call site.
- `CountsProvider` / `TermsProvider`: added unmount-cleanup `useEffect` clearing `timerRef` so debounced flushes don't fire after the page navigates away.

src/containers/diseasePage/index.jsx
```diff
@@ -19,7 +19,7 @@ const SUMMARY = 'Summary';
 const GENES = 'Associated Genes';
 const ALLELES = 'Associated Alleles';
 const MODELS = 'Associated Models';
-const SECTIONS = [{ name: SUMMARY }, { name: GENES }, { name: ALLELES }, { name: MODELS }];
+const BROWSE_ONTOLOGY = 'Browse Ontology';
```
sethrcamp, line 22: This may be more of a discussion topic than feedback, but I don't understand the pattern we have established throughout the project of using constants in this way. They are only referenced twice, and I feel like using string literals would be easier to read in context.

Since this is all passed to the `<DataPage/>` component as children, it would make sense to me that you should only need to provide the information on which sections are on the page 1 time. I suppose doing this way allows you to remove sections from the side navigation, but that could also easily be done with an option flag.

Just some food for thought

src/containers/ontologyBrowser/annotationTypes.js
```diff
@@ -0,0 +1,30 @@
+// Annotation type configuration for ontology-browser count pills.
```
sethrcamp, line 1: 2 things:

1) Is there a reason for the double nested directories (i.e. `ontologyBrowser/browse/fileName.js` instead of simply `ontologyBrowser/fileName.js`?

2) This would be a good opportunity for TypeScript to prevent invalid configurations of new AnnotationTypes if any are added in the future

src/containers/ontologyBrowser/browse/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+import { TermsProvider } from './useDiseaseTerms.jsx';
+import style from './style.module.scss';
+
+const PAGE_TITLE = 'Ontology Browser';
+
+const OntologyBrowser = () => {
+  const { ontology: ontologyParam, id } = useParams();
+  const navigate = useNavigate();
+  const location = useLocation();
+  const ontology = getOntology(ontologyParam || DEFAULT_ONTOLOGY_ID);
```
sethrcamp: Having the `|| DEFAULT_ONTOLOGY_ID` is redundant since the `getOntology` function already handles a default case.

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+  const detailCurie = useMemo(() => focusedCurie || ontology.rootCurie || null, [focusedCurie, ontology.rootCurie]);
+
+  return (
+    <CountsProvider>
+      <TermsProvider>
+        <div className={style.page}>
+          <HeadMetaTags title={PAGE_TITLE} />
+          <h2>{PAGE_TITLE}</h2>
+
+          <div className={style.chooserRow}>
```
sethrcamp, line 64: I think an improvement here would be to combine the dropdown/search box like we do in the search bar in the header. It would condense the UI and be a recognizable pattern to the user.

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+              {ONTOLOGIES.map((o) => (
+                <option key={o.id} value={o.id}>
+                  {o.label}
+                  {o.available ? '' : ' — coming soon'}
+                </option>
+              ))}
+            </select>
+          </div>
+
+          {!ontology.available ? (
```
sethrcamp, line 83: You should be able to disable the "coming soon" options within the select dropdown, which would prevent the need for this text

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+            <div className={style.placeholder}>
+              <p>
+                <strong>{ontology.label}</strong> is not yet available in the browser. This ontology has not been
+                indexed into Elasticsearch. Pick another ontology above, or check back once it is loaded.
+              </p>
+            </div>
+          ) : (
+            <>
+              <p style={{ color: '#6c757d' }}>
```
sethrcamp: I'm not sure we need this explainer text. I think the UI is intuitive enough without it

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+            </label>
+            <select
+              id="ontology-chooser"
+              className="form-control"
+              style={{ maxWidth: 360, display: 'inline-block' }}
+              value={ontology.id}
+              onChange={(e) => handleOntologyChange(e.target.value)}
+            >
+              {ONTOLOGIES.map((o) => (
+                <option key={o.id} value={o.id}>
```
sethrcamp, line 75: The key here is technically not specific enough. Although it would likely work, it could break if this component were on the page more than once, or if an ontology were accidentally in the list more than once. Ideally, a key should be truly unique. The pattern I typically use is:
```
const id = useId();
...
(o,index) => (
<option key={`${id}-ontology-dropdown-${o.id}-${index}`} ...
```
The `useId` hook spits out an unique id, which is different for each instance of the component. Adding "ontology-dropdown" ensures uniqueness from anywhere else in the component where you may be looping through the same list of ontologies. The combination of `o.id` and `index` ensures that a repeat ontology id is still unique because of its index. Technically with this pattern you don't need the ontology id at all, but I have observed it still being used as a carry over from when the `useId` hook didn't exist yet.

src/containers/ontologyBrowser/browse/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+              id="ontology-chooser"
+              className="form-control"
+              style={{ maxWidth: 360, display: 'inline-block' }}
+              value={ontology.id}
+              onChange={(e) => handleOntologyChange(e.target.value)}
+            >
+              {ONTOLOGIES.map((o) => (
+                <option key={o.id} value={o.id}>
+                  {o.label}
+                  {o.available ? '' : ' — coming soon'}
```
sethrcamp: Can be simplified to `{o.available && " - coming soon"}`

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+          <h2>{PAGE_TITLE}</h2>
+
+          <div className={style.chooserRow}>
+            <label htmlFor="ontology-chooser" className={style.chooserLabel}>
+              Ontology:
+            </label>
+            <select
+              id="ontology-chooser"
+              className="form-control"
+              style={{ maxWidth: 360, display: 'inline-block' }}
```
sethrcamp: Unless there is a reason not to, I would shift this to the scss file

src/containers/ontologyBrowser/browse/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+          <div className={style.chooserRow}>
+            <label htmlFor="ontology-chooser" className={style.chooserLabel}>
+              Ontology:
+            </label>
+            <select
+              id="ontology-chooser"
+              className="form-control"
+              style={{ maxWidth: 360, display: 'inline-block' }}
+              value={ontology.id}
+              onChange={(e) => handleOntologyChange(e.target.value)}
```
sethrcamp: Since this is only used once, you could swap for ```onChange={(e) => navigate(`/ontology/${e.target.value}`)}```

src/containers/ontologyBrowser/ontologies.js
```diff
@@ -0,0 +1,24 @@
+// Supported ontologies in the browser. Only entries with `available: true` are
+// backed by data today; the others render a "coming soon" placeholder until
+// they are indexed into Elasticsearch.
+export const ONTOLOGIES = [
```
sethrcamp, line 4: Same comment as annotationTypes.js

src/containers/ontologyBrowser/browse/OntologySearchBox.jsx
```diff
@@ -0,0 +1,219 @@
...
+            </div>
+          );
+        }}
+        inputProps={{
+          value,
+          placeholder: placeholder || 'Search terms',
+          onChange: (_e, { newValue }) => setValue(newValue),
+          className: 'form-control',
+        }}
+        theme={{
```
sethrcamp: I would split large static literals like this out into a variable, outside of the component to improve readablility

src/containers/ontologyBrowser/OntologySearchBox.jsx
```diff
@@ -0,0 +1,219 @@
...
+              <div style={{ maxHeight: 280, overflowY: 'auto' }}>{children}</div>
+              {children && query && (
+                <button
+                  type="button"
+                  aria-label={`View all results for ${query}`}
+                  onMouseDown={(e) => {
+                    e.preventDefault();
+                    openModalFor(query);
+                  }}
+                  style={{
```
sethrcamp: I would put this in a scss file

src/containers/ontologyBrowser/OntologySearchBox.jsx
```diff
@@ -0,0 +1,219 @@
...
+          <div style={{ padding: '4px 8px' }}>
+            <strong>{s.name || s.nameKey}</strong>{' '}
+            <span style={{ color: '#868e96', fontSize: '0.8rem', fontFamily: 'monospace' }}>{s.curie}</span>
+          </div>
+        )}
+        renderSuggestionsContainer={({ containerProps, children, query }) => {
+          const { key, ...containerRest } = containerProps;
+          return (
+            <div key={key} {...containerRest}>
+              <div style={{ maxHeight: 280, overflowY: 'auto' }}>{children}</div>
```
sethrcamp: I would put this in an scss file

src/containers/ontologyBrowser/index.jsx
```diff
@@ -0,0 +1,139 @@
...
+                  onSelect={handleSelect}
+                  category={ontology.searchCategory}
+                  placeholder={`Search ${ontology.label} terms`}
+                />
+              </div>
+
+              <div className={style.legend}>
+                <span className={style.legendLabel}>Annotation Available:</span>
+                {ANNOTATION_TYPES.map((t) => (
+                  <span key={t.id} className={style.legendItem} style={{ background: t.bg, color: t.fg }}>
```
sethrcamp, line 108: Same comment as above with key.

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+  curie: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  depth: PropTypes.number.isRequired,
+  forceExpanded: PropTypes.instanceOf(Set).isRequired,
+  focusedCurie: PropTypes.string,
+  onSelect: PropTypes.func.isRequired,
+  scrollOnFocus: PropTypes.bool,
+};
+
+const OntologyTree = ({ rootCurie, rootName, forceExpanded, focusedCurie, onSelect, scrollOnFocus = true }) => {
```
sethrcamp: Although I understand thinking of a tree having nodes as children, this structure is more similar to a tree having trees as children. I think think this wrapper implies that you can only make an ontology tree by starting at a root ontology term, but there is no reason you couldn't start with a term that has a parent term, and only render everything below it

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
+import React, { useEffect, useState, useRef } from 'react';
+import PropTypes from 'prop-types';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faChevronRight, faChevronDown } from '@fortawesome/free-solid-svg-icons';
+import CountBadge from './CountBadge.jsx';
+import { ANNOTATION_TYPES } from './annotationTypes.js';
+import { useDiseaseTerm } from './useDiseaseTerms.jsx';
+import style from './style.module.scss';
+
+const TreeNode = ({ curie, name, depth, forceExpanded, focusedCurie, onSelect, scrollOnFocus }) => {
```
sethrcamp: Going off of my below comment, I would convert this whole file to a single component called "OntologyTree". I think "TreeNode" is too abstract for a component that specifically works for ontologies only

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+  const [open, setOpen] = useState(false);
+  const rowRef = useRef(null);
+
+  // Term doc comes from the batched provider so every visible node knows its
+  // children without firing per-node requests.
+  const data = useDiseaseTerm(curie);
+
+  useEffect(() => {
+    if (forceExpanded.has(curie)) setOpen(true);
+  }, [forceExpanded, curie]);
```
sethrcamp, line 21: `setOpen` is missing from the list of dependancies

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  useEffect(() => {
+    if (scrollOnFocus && focusedCurie === curie && rowRef.current) {
+      rowRef.current.scrollIntoView({ block: 'center', behavior: 'smooth' });
+    }
+  }, [focusedCurie, curie, scrollOnFocus]);
+
+  const children = data?.children || [];
+  // We don't know if a node has children until we've expanded it and
```
sethrcamp, line 30: I'll include this in a longer comment/discussion at a UI meeting, but it would be a better user experience to always have the children loaded fully, so that they are automatically there when the dropdown is toggled. This comes with some challenges of its own, but I think they would be worth it

src/containers/ontologyBrowser/browse/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+    if (scrollOnFocus && focusedCurie === curie && rowRef.current) {
+      rowRef.current.scrollIntoView({ block: 'center', behavior: 'smooth' });
+    }
+  }, [focusedCurie, curie, scrollOnFocus]);
+
+  const children = data?.children || [];
+  // We don't know if a node has children until we've expanded it and
+  // fetched its doc. Show the arrow optimistically; hide it only after a
+  // fetch confirms no children.
+  const knownEmpty = !!data && children.length === 0 && !(data.doTerm?.descendantCount > 0);
```
sethrcamp: Could be simplified to `const knownEmpty = children.length === 0 && data?.doTerm?.descendantCount === 0;`

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+  // fetched its doc. Show the arrow optimistically; hide it only after a
+  // fetch confirms no children.
+  const knownEmpty = !!data && children.length === 0 && !(data.doTerm?.descendantCount > 0);
+  const hasChildren = !knownEmpty;
+  const isFocused = focusedCurie === curie;
+
+  const toggle = (e) => {
+    e.stopPropagation();
+    setOpen((v) => !v);
```
sethrcamp, line 39: Could be simplified to `setOpen(!open)`

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+            type="button"
+            className={style.toggle}
+            onClick={toggle}
+            title={open ? 'Collapse' : 'Expand children'}
+            aria-label={open ? 'Collapse' : 'Expand children'}
+          >
+            <FontAwesomeIcon icon={open ? faChevronDown : faChevronRight} fixedWidth />
+          </button>
+        ) : (
+          <span className={style.toggleSpacer} />
```
sethrcamp, line 60: You could avoid this spacer span with styling or a wrapper element

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+        )}
+        <span>{name}</span>
+        <span className={style.curie}>&nbsp;{curie}</span>
+        <span className={style.badgeRow}>
+          {ANNOTATION_TYPES.map((t) => (
+            <CountBadge key={t.id} curie={curie} type={t} />
```
sethrcamp, line 66: Same comment about the key

src/containers/ontologyBrowser/browse/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+    if (forceExpanded.has(curie)) setOpen(true);
+  }, [forceExpanded, curie]);
+
+  useEffect(() => {
+    if (scrollOnFocus && focusedCurie === curie && rowRef.current) {
+      rowRef.current.scrollIntoView({ block: 'center', behavior: 'smooth' });
+    }
+  }, [focusedCurie, curie, scrollOnFocus]);
+
+  const children = data?.children || [];
```
sethrcamp: I would rename this to something like "childTerms" or something similar. "children" is typically reserved to the prop of the same name

src/containers/ontologyBrowser/browse/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+  const toggle = (e) => {
+    e.stopPropagation();
+    setOpen((v) => !v);
+  };
+
+  return (
+    <div className={style.node}>
+      <div
+        ref={rowRef}
+        className={`${style.nodeRow} ${isFocused ? style.nodeRowFocused : ''}`}
```
sethrcamp: `className={classNames(style.nodeRow, { [style.nodeRowFocused]: isFocused })}`

I love this library for cases exactly like then, and especially so if you have multiple conditional class names

src/containers/ontologyBrowser/OntologyTree.jsx
```diff
@@ -0,0 +1,124 @@
...
+            title={open ? 'Collapse' : 'Expand children'}
+            aria-label={open ? 'Collapse' : 'Expand children'}
+          >
+            <FontAwesomeIcon icon={open ? faChevronDown : faChevronRight} fixedWidth />
+          </button>
+        ) : (
+          <span className={style.toggleSpacer} />
+        )}
+        <span>{name}</span>
+        <span className={style.curie}>&nbsp;{curie}</span>
```
sethrcamp, line 63: This is a matter of preference, but you can also do `{" "}{curie}`

src/containers/ontologyBrowser/style.module.scss
```diff
@@ -0,0 +1,245 @@
...
+  max-width: 1520px;
+  margin: 0 auto;
+}
+
+.layout {
+  display: flex;
+  gap: 1.25rem;
+  flex-direction: column;
+
+  @media (min-width: 768px) {
```
sethrcamp, line 17: I would avoid setting breakpoints if at all possible.

Thinking about the mobile experience of this browser, I think it would be better in general if we had no wrapping at all in the tree view. Then the side-by-side nature could be dealt with using flex wrapping and min-widths

src/containers/ontologyBrowser/browse/style.module.scss
```diff
@@ -0,0 +1,245 @@
+@import '../../../theme.module';
```
sethrcamp: I would really recommend utilizing scss's nesting ability. I know with modules it can get quite cumbersome to use classnames all the way down, but personal I find this:
```
.myList {...}
.myLIstItem {...}
.myLIstItemContent {...}
```
a lot harder to understand than this:
```
.myList {
  ...
  li {
    ...
    > div {
      ...
    }
  } 
}

File: src/containers/ontologyBrowser/style.module.scss [hunk 29]
```diff
@@ -0,0 +1,245 @@
+@import '../../../theme.module';
+@import '~bootstrap/scss/functions';
```
sethrcamp, line 2: I would also recommend not using bootstrap. Bootstrap 4 is going to be EOLed in the not too distant future, and the upgrade path to Bootstrap 5 is a nightmare. The less we have to change at that point, the better

oblodgett (approved):

src/containers/ontologyBrowser/TermDetailPanel.jsx
```diff
@@ -0,0 +1,110 @@
...
+  const parents = data.parents || [];
+  const portalSlug = portalSlugForCurie(curie);
+
+  return (
+    <div>
+      <h3 className={style.detailTitle}>{term.name}</h3>
+      <div className={style.curie}>{curie}</div>
+      <div className={style.detailBadgeRow}>
+        {ANNOTATION_TYPES.map((t) => (
+          <CountBadge key={t.id} curie={curie} type={t} expanded />
```
sethrcamp, line 43: Same comment about key

src/containers/ontologyBrowser/TermDetailPanel.jsx
```diff
@@ -0,0 +1,110 @@
...
+          </ul>
+        </div>
+      )}
+
+      {synonyms.length > 0 && (
+        <div className={style.detailSection}>
+          <div className={style.detailSectionLabel}>Synonyms</div>
+          <CollapsibleList collapsedSize={5}>
+            {synonyms.map((s, i) => (
+              <span key={i}>{typeof s === 'string' ? s : s?.name || JSON.stringify(s)}</span>
```
sethrcamp, line 84: Same comment about key

src/containers/ontologyBrowser/TermDetailPanel.jsx
```diff
@@ -0,0 +1,110 @@
...
+          </CollapsibleList>
+        </div>
+      )}
+
+      {xrefs.length > 0 && (
+        <div className={style.detailSection}>
+          <div className={style.detailSectionLabel}>Cross-references</div>
+          <CollapsibleList collapsedSize={5}>
+            {xrefs.map((x, i) => (
+              <ExternalLink key={i} href={x.url}>
```
sethrcamp, line 95: Same comment about key

src/containers/ontologyBrowser/useDiseaseCounts.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { createContext, useCallback, useContext, useEffect, useRef, useState } from 'react';
```
sethrcamp, line 1: In general I think the counts should come back as apart of the same API call as the other info

src/containers/ontologyBrowser/useDiseaseTerms.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const timerRef = useRef(null);
+  const termsRef = useRef(terms);
+
+  useEffect(() => {
+    termsRef.current = terms;
+  }, [terms]);
+
+  useEffect(() => {
+    return () => {
+      if (timerRef.current) clearTimeout(timerRef.current);
```
sethrcamp: Same comment about conditional for refs

src/containers/diseasePage/index.jsx
jdepons: This file uses the same `const SECTION_NAME = '...'` pattern as genePage, allelePage, referencePage, and the other DataPage callers pre-existing pages, not added in this PR. I matched the existing convention rather than diverge in one file. Happy to discuss at UI meeting.

src/containers/ontologyBrowser/annotationTypes.js
jdepons: First part is done and pushed in 25adf753.

For the second part — TypeScript is set up for infra (cdk/) and build tooling (vite.config.ts, jest.config.ts), but no `src/` code uses it yet. Personally I'd rather not adopt it 😄, and even if we did, starting with a one-off file here seems like the wrong way in.  It feels like a bigger discussion to make sure everyone's on board first.

src/containers/ontologyBrowser/index.jsx
jdepons: The two controls do different things here — the chooser navigate to a different ontology URL on change, while the search returns  a term within the current ontology (with a "view all results" modal underneath). Combining them like the header search bar would either change the URL scheme (no longer one URL per ontology) or end up as a navigate-on-change control inside a search input-group, which isn't really the same pattern.                                                
                                                                    
Happy to revisit if we want to redesign the chooser as a scope filter rather than navigation, but treating that as a separate discussion / ticket.

src/containers/ontologyBrowser/index.jsx
jdepons: The coming soon options are there to point out we are missing in the index.  These will not go to production.

src/containers/ontologyBrowser/index.jsx
jdepons: I'll run this by the user group.   I don't feel strongly about it either way.    We can see what the end users want.

src/containers/ontologyBrowser/index.jsx
jdepons: `o.id` from `ONTOLOGIES` is already a stable unique value, and React keys only need to be unique among siblings (not globally), so a second instance of this component on the page wouldn't collide.

src/containers/ontologyBrowser/ontologies.js
jdepons: See reply on annotationTypes.js.  Same answer on both points.

src/containers/ontologyBrowser/index.jsx
jdepons: See reply on index.jsx:77.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: `setOpen` is a React `useState` setter, which has a stable identity guaranteed by React and is whitelisted by the `react-hooks/exhaustive-deps` rule, so omitting it from the dependency array is correct.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: Sounds good.  I am happy to revisit at the UI meeting.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: Keeping the functional updater setOpen((v) => !v).  It reads the latest state inside the setter rather than capturing open from the closure, so it stays correct if the toggle ever fires  from a stale callback or batched update.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: The spacer keeps the toggle column aligned between rows with and without children. Could move to CSS padding + an absolutely-positioned button instead functionally equivalent, just trades a span for a positioning rule. Leaving as-is.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: See reply on index.jsx:77.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: Keeping &nbsp; so the curie stays on the same line as the name (non-breaking).

src/containers/ontologyBrowser/style.module.scss
jdepons: Treating this as a styling follow-up, since the breakpoint matches Bootstrap's md convention used elsewhere and the tree-row wrapping question is related. Can revisit together.

src/containers/ontologyBrowser/style.module.scss
jdepons: Valid concern, but it is a project-wide direction (bootstrap 4.6.1 is pinned and used pervasively).

src/containers/ontologyBrowser/TermDetailPanel.jsx
jdepons: See reply on index.jsx:77.

src/containers/ontologyBrowser/TermDetailPanel.jsx
jdepons: See reply on index.jsx:77.

src/containers/ontologyBrowser/TermDetailPanel.jsx
jdepons: See reply on index.jsx:77.

src/containers/ontologyBrowser/useDiseaseCounts.jsx
jdepons: The split was intentional. Term lookups are a fast ES doc fetch that fires per visible node, while counts are heavier aggregation queries, so combining them would slow the tree even when counts are hidden. Worth revisiting as a backend follow-up.

src/containers/ontologyBrowser/browse/index.jsx
jdepons: Dropped in acd4a659. getOntology already falls back to the default.

src/containers/ontologyBrowser/browse/index.jsx
jdepons: Done in acd4a659. Simplified to `{!o.available && ' — coming soon'}`.

src/containers/ontologyBrowser/index.jsx
jdepons: Done in e1e57910. Moved to a `.chooserSelect` class in style.module.scss.

src/containers/ontologyBrowser/browse/index.jsx
jdepons: Done in acd4a659. Inlined as `onChange={(e) => navigate(...)}` and dropped the named handler.

src/containers/ontologyBrowser/browse/OntologySearchBox.jsx
jdepons: Done in acd4a659. Hoisted to a module constant AUTOSUGGEST_THEME.

src/containers/ontologyBrowser/OntologySearchBox.jsx
jdepons: Done in e1e57910. Moved to a `.viewAllButton` class in style.module.scss.

src/containers/ontologyBrowser/OntologySearchBox.jsx
jdepons: Done in e1e57910. Moved to a `.suggestionsScroll` class in style.module.scss.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: Done in f8383ae1. Dropped the wrapper, renamed TreeNode to OntologyTree, and changed the caller's rootCurie/rootName props to curie/name. Now mounts from any term, not just a root.

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: Done in f8383ae1. The file is now a single OntologyTree component.

src/containers/ontologyBrowser/browse/OntologyTree.jsx
jdepons: The proposed `data?.doTerm?.descendantCount === 0` is subtly different. When doTerm is missing entirely it evaluates false (arrow visible), while the current code evaluates true (arrow hidden). Keeping current behavior.

src/containers/ontologyBrowser/browse/OntologyTree.jsx
jdepons: Done in acd4a659. Renamed `children` to `childTerms`.

src/containers/ontologyBrowser/browse/OntologyTree.jsx
jdepons: Done in acd4a659. Switched to `classNames(style.nodeRow, { [style.nodeRowFocused]: isFocused })`.

src/containers/ontologyBrowser/browse/style.module.scss
jdepons: Noted. Will adopt the nested style in the upcoming styling pass.

src/containers/ontologyBrowser/useDiseaseTerms.jsx
jdepons: Dropped in 85bc270d at all four sites in useDiseaseTerms.jsx

Joe-Wdz (approved):
Seth's comments pretty much covered everything.

src/containers/ontologyBrowser/annotationTypes.js
sethrcamp: Fair enough. The upcoming table replacement uses it heavily, because it was adapted from code I had already wrote for FlyBase. It was going to serve as the first experiment in TypeScript for the repo, but has been held up for a while

src/containers/ontologyBrowser/index.jsx
sethrcamp: I don't think we need any sort of functional change here, I was speaking on visuals alone. It may not be a combined thing under the hood, which is fine, but to the user it would look like they were combined. From the user perspective, if they look like they are the same, then intuition leads them to think "Ah, this must search for a term within the selected ontology", without having to explicitly state it on the page.

I always like to go off of the rule of thumb for web design that states "if you have to tell the user how to use the thing, then there is probably a more intuitive way to design it". (I.e. instead of some text saying "click _here_ to write and send an email to the president", have a call-to-action button that says "email the president")

src/containers/ontologyBrowser/index.jsx
sethrcamp: Ah, if they aren't intended to appear in the drop down at all on production, then they should probably be removed before merging to stage.

I don't personally see a problem with having them as non-selectable options in the dropdown (although that would make the dropdown useless until we add a second ontology), but other folks may have a different preference

src/containers/ontologyBrowser/index.jsx
sethrcamp: Huh, I'm not sure when it changed (because I probably haven't looked into it for the last 10 years), but React does now in fact scope elements being mapped. I wonder how long I have been needlessly making very unique keys. 

I'm happy to have learned this! Bug squashing was a giant headache several years ago when you had the rare event that React got confused with its keys

src/containers/ontologyBrowser/OntologyTree.jsx
sethrcamp: Now I feel like a broken record, but I wonder how long I have been needlessly adding these functions to dependency arrays. I remember starting out with functional components and almost always forgetting them and having ESLint yell at me. I'm probably years late to the party, but celebrating anyways!

src/containers/ontologyBrowser/OntologyTree.jsx
sethrcamp: The reason I would solve this with styling or a wrapper element is because it is better for screen readers to not have empty elements, especially since it is a span, which screen readers tend to treat as containing text-based content.

Styling would remove the empty element entirely, and a wrapper would make it clear while using a screen reader that there may or may not be a toggle in that location, whereas that context is lost in the current implementation since there will either be a span or a button, but consistently one element

src/containers/ontologyBrowser/style.module.scss
sethrcamp: That is true, however we discovered the bootstrap problem while upgrading to vite, and it is already giving us some mild issues (for example, the mountain of scss errors/warnings when running the project). It is truly going to be a nightmare to fix in the future, and I would love to not make the nightmare even bigger.

Also, there are already several different styling paradigms in the project, so I am not too worried about consistency (at least until we resolve the bootstrap issue)

src/containers/ontologyBrowser/useDiseaseCounts.jsx
sethrcamp: This might be a good opportunity to add directly in the ES documents. That would be an indexer problem though, so not much we could do there for now. Counts are usually quite fast within postgres, so I would imagine it wouldn't be too taxing for the indexer as compared to ES

src/containers/ontologyBrowser/useDiseaseTerms.jsx
sethrcamp: I could be wrong, but I think this is still susceptible to the same issue, though I don't think it will ever happen in these cases as-is.

clearTimeout may be okay with accepting null/undefined, however, if `timerRef` is null, then an error will be thrown trying to access the "current" attribute of "null". 

In the cleanup function I wouldn't expect this to ever happen, and based on the logic of the two useCallbacks, I wouldn't expect it to happen there either

It is probably okay to leave as-is, but if you ever use a ref to attach to an element, you will almost certainly have to put in the `fooRef && fooRef.current` or `fooRef?.current` check

src/containers/ontologyBrowser/OntologyTree.jsx
jdepons: We can chat about it at the UI meeting

PR #1774, opened 2026-05-20
Feature/automated mergebacks

sethrcamp:
Adapted from the automated mergeback system of the curation repo.

The biggest difference is the omission of the tagging system & requiring a version number for the branch name. Although I do like using semantic versioning, the rules laid out for the public site version do not account for a "code but not data" update to the version number. Even still, making an update to the actual site version number displayed on the public site takes additional steps typically done within the full release pipeline.


Joe-Wdz:
It looks good based on Claude Code’s review

Joe-Wdz (approved):

PR #1784, opened 2026-06-03
KANBAN-1056: Disease Portal "Definition" → "Summary" with citations and synonyms

jimflybase:
## Summary

Implements KANBAN-1056: redesigns the Disease Portal detail-page **"Definition"** section as a **"Summary"** section, modeled on the first two parts of the Disease page Summary.

- Rename the portal section heading and left-nav label from **Definition** → **Summary**.
- New `SummarySection` renders the DO term **definition** with inline **citation links** (`definitionUrls`, each `[n]` with a URL hover tooltip) and **synonyms** — reusing the same shared components as the Disease page (`ExternalLink`, `CommaSeparatedList`, `SynonymList`, `formatLink`).
- Adds small uppercase **DEFINITION** / **SYNONYMS** sub-head labels styled for the portal look.

No new data fetching — the portal page already fetched `/api/disease/{doid}`; the citations and synonyms were in that payload, just unused.

## Incidental fixes (found during verification)

- **PapersSection hydration error:** floated species-sprite `<div>`s were nested inside `<p>` (invalid HTML → React hydration warning). Switched the publication wrappers to `<div>` and restored inter-entry spacing via a `.publicationEntry` class. Console errors on portal pages went from 2 → 0.
- **Trailing whitespace below the Members banner:** the shared full-width banner's chrome (border + padding) plus the wrapping subsection's 4rem margin left a large gap above the footer. A scoped `.membersFooter` wrapper trims it on detail pages; the list-page banner is unaffected.

## Testing

Verified in the browser across all three detail portals — **Alzheimer's**, **Parkinson's**, **Diabetes Mellitus**:
- Definition, citation links (valid hrefs, `target="_blank"`, working URL tooltips), and synonyms render correctly, handling varying counts.
- 0 console errors (Parkinson's has no publications and was already clean — confirming the errors were from PapersSection, not the Summary work).
- Members banner now sits cleanly above the footer; list-page banner unchanged.

### DIFF
--- a/src/containers/diseasePortal/PapersSection.jsx
+++ b/src/containers/diseasePortal/PapersSection.jsx
@@ -5,6 +5,7 @@ import SpeciesIcon from '../../components/speciesIcon/index.jsx';
 import { speciesMap } from '../referencePage/index.jsx';
 import { Link } from 'react-router-dom';
 import usePageLoadingQuery from '../../hooks/usePageLoadingQuery';
+import style from './style.module.scss';
 
 const CitationLink = ({ curie }) => {
   const { data: pubData, isLoading, isError } = usePageLoadingQuery(`/api/reference/${curie}`);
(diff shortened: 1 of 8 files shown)


jimflybase:
Thanks for the catch — fixed in ddec58b4.

- Hoisted `usePageLoadingQuery` above the conditional early returns in `DiseasePortalPage` so the hook count is constant across the list/detail routes (which share a fiber).
- Passes a `null` url when there's no `doid`, and added `enabled: Boolean(url)` to `usePageLoadingQuery` so a null url is a safe no-op rather than fetching an undefined endpoint.

Verified in-browser: list → detail, back → list, and list → other-detail navigations now all render with 0 console errors (previously this white-screened with "Rendered more hooks than during the previous render").

The non-blocking note about `PortalListSection` firing 3 count requests per item is acknowledged — left as-is for now since it matches the existing `DiseasePortalSection` pattern.

jdepons (approved):
Tough for me to review react code.  That said, nothing stands out to me an incorrect.

PR #1795, opened 2026-06-17
Release/kanban 461 462 1084

Joe-Wdz:
KANBAN-461: Add 'Your Input Welcome' button
KANBAN-462: Create 'contact us' form linked to 'Contact Us' menu item
KANBAN-1084: Disease ribbon cell color for single annotations


src/constants.js
```diff
@@ -1,4 +1,6 @@
+export const CONTACT_API_URL = 'https://1ythblxki4.execute-api.us-east-1.amazonaws.com/prod/contact';
 export const HELP_EMAIL = 'help@alliancegenome.org';
+export const RECAPTCHA_SITE_KEY = '6Ld-r5EsAAAAAMlCU84nH_F5o4_sLkJYnBhr3j8b';
```
sethrcamp: This should probably be stored as a secret in the repo or an env variable. Chris T should be able to give you permissions to do so if you don't have them already

sethrcamp (approved):

src/constants.js
Joe-Wdz: It is not a secret and will be downloaded by every browser. but for management i agree that we can move it to env or repo. i will ask Chris T to assist.  thanks.

PR #1801, opened 2026-06-26
feat: embed scoped disease ontology browser on portal pages (KANBAN-1391)

jimflybase:
## Summary
Embeds a scoped Disease Ontology browser on disease portal detail pages (KANBAN-1391). Adds a new **Ontology View** section at the bottom of the portal page showing an in-page ontology tree rooted at the focus term, with drill-down into descendants and links out to the full-page ontology browser.

## Changes
- **New `OntologyContextSection.jsx`**: renders the focus term's immediate parents as links to the standalone ontology browser, plus an in-page tree rooted at the focus term with expand/collapse drill-down. Reuses the existing `ontologyBrowser` tree and batch providers.
- **`OntologyTree.jsx`**: generalized the leaf-only `leafHref` into a `nodeHref` applied to *all* nodes, so every term links to the browser focused at that term (drill-down stays on the chevron). Also natural-sorts numbered sibling terms (`localeCompare` with `{ numeric: true }`) so "disease 1, 2, 3, 14" order numerically.
- **`index.jsx`**: registers the **Ontology View** section + sidebar jump-link.
- **`style.module.scss`**: styles for the section, parent-term links, and the G|M|A "Annotation Available" legend.

## Notes
- Multi-parent (DAG) terms don't focus correctly in the standalone browser when opened via these links — tracked separately in **KANBAN-1395**.
- Branch includes a merge of `stage` (brings in the recently merged Recent Papers work); the net PR diff is only the KANBAN-1391 ontology files.

### DIFF
--- a/src/containers/diseasePortal/OntologyContextSection.jsx
+++ b/src/containers/diseasePortal/OntologyContextSection.jsx
@@ -1,0 +1,74 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Link } from 'react-router-dom';
+import { useQuery } from '@tanstack/react-query';
+import fetchData from '../../lib/fetchData';
+import OntologyTree from '../ontologyBrowser/OntologyTree.jsx';
+import { CountsProvider } from '../ontologyBrowser/useDiseaseCounts.jsx';
+import { TermsProvider } from '../ontologyBrowser/useDiseaseTerms.jsx';
+import style from './style.module.scss';
+
+// Embedded, scoped ontology view for a disease portal page (KANBAN-1391).
+// Two interaction modes by region:
+//  - Ancestor context (immediate parents) renders as links OUT to the full-page
+//    ontology browser, so the user can jump up the hierarchy.
+//  - The focus term and its descendants stay in-page: the reused OntologyTree
+//    handles expand/collapse drill-down locally, no route change.
+const OntologyContextSection = ({ curie, name }) => {
+  // Immediate parents come from the single-term doc (same shape the standalone
+  // TermDetailPanel uses). The DO is a DAG, so there can be more than one.
+  const { data, isLoading } = useQuery({
+    queryKey: ['disease-term', curie],
+    queryFn: () => fetchData(`/api/disease/${encodeURIComponent(curie)}`),
+    enabled: !!curie,
+    staleTime: 5 * 60_000,
+  });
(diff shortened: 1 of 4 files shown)


jimflybase:
**@jdepons** — reviewer note as the ontology-browser stakeholder. This PR **modifies the shared `OntologyTree` component**, not just embeds it. Three things, in priority order:

**1. One change affects the standalone browser globally (your sign-off matters here):**
The sibling sort now uses `localeCompare(…, undefined, { numeric: true })`. This is **not opt-in** — it changes ordering everywhere `OntologyTree` renders, including the standalone browser. Numbered siblings now sort numerically (`disease 1, 2, 3, 14`) instead of lexicographically (`1, 14, 17, 2, 3`). Almost certainly an improvement, but it's the one behavior change you'll see in your own tool, so it's the thing to actively confirm you're happy with.

**2. The `nodeHref` addition is backward-compatible (FYI, no action):**
New optional `nodeHref` prop makes every term label a `<Link>`. The standalone browser doesn't pass it → falls through to the existing `<span>` → **renders identically to today**. Label clicks `stopPropagation()` so linking never interferes with chevron drill-down. Only the portal embed opts in (pointing at `/ontology/disease/<curie>`).

**3. A pre-existing browser limitation this surfaces — KANBAN-1395:**
Multi-parent (DAG) terms don't focus correctly **in the standalone browser** when opened via the embed's deep links. This isn't a regression introduced here — it's an existing focus limitation the new links simply expose. Filed separately as KANBAN-1395.

Also: the embed reuses your batch providers (`useDiseaseTerm`/`useDiseaseTerms`) and the tree as-is — no provider changes — so the data-fetching path you own is unchanged.

jimflybase:
Fixed in 054672a6. Confirmed the bug: the portal page reuses its fiber across `/disease-portal/:name` navigations, and `OntologyContextSection` seeds both `selectedCurie` and `forceExpanded` from `curie` via `useState`, so they went stale on navigation — the new focus term no longer auto-expanded and lost its highlight.

Went with the **remount** option (`key={diseaseData.doid}`) rather than reactive derivation: keying the component also resets the inner `OntologyTree` rows' `open` state, so a new disease gets a fully fresh, correctly-scoped tree. Reactive derivation in the parent would have left those inner row states stale.

jdepons (approved):

PR #1815, opened 2026-07-09
KANBAN-715 convert trivial class components

Joe-Wdz:


src/components/dataTable/columnHeader.jsx
```diff
@@ -7,44 +7,41 @@ import { faFilter } from '@fortawesome/free-solid-svg-icons';
...
-    const { column, filter, filterElement } = this.props;
-    const classes = '';
-    const popperModifiers = [
-      {
-        name: 'preventOverflow',
-        options: {
-          rootBoundary: 'viewport',
-        },
+const ColumnHeader = ({ column, filter, filterElement }) => {
+  const classes = '';
```
sethrcamp: I think this variable is unnecessary. It never gets changed and is only passed as a className, which has no effect as an empty string.

src/components/dataTable/columnHeader.jsx
```diff
@@ -7,44 +7,41 @@ import { faFilter } from '@fortawesome/free-solid-svg-icons';
...
+  const popperModifiers = [
```
sethrcamp: We should move this outside of the component since it isn't statefull

src/components/dataTable/columnHeader.jsx
```diff
@@ -7,44 +7,41 @@ import { faFilter } from '@fortawesome/free-solid-svg-icons';
...
+        <UncontrolledButtonDropdown>
+          <DropdownToggle className={`${style.filterToggle} ${active ? style.active : ''}`} color="link" tag="span">
+            <FontAwesomeIcon icon={faFilter} />
+          </DropdownToggle>
+          <DropdownMenu className="shadow-sm px-4 py-3" modifiers={popperModifiers} strategy="fixed">
+            {filterElement}
+          </DropdownMenu>
+        </UncontrolledButtonDropdown>
+      )}
+      {helpPopupProps ? (
```
sethrcamp: `helpPopupProps &&`

src/components/subsection.jsx
```diff
@@ -10,60 +10,54 @@ import NoData from './noData.jsx';
...
-    const id = this.props.title && makeId(this.props.title);
-    const target = <a className={style.target} id={id} />;
-    const helpPopup = this.props.help && (
-      <span className="small ml-3">
-        <HelpPopup id={`help-${this.props.title}`}>{this.props.help}</HelpPopup>
-      </span>
-    );
+const Subsection = ({ children, hardcoded, hasData, help, hideTitle = false, isMeta, level, title }) => {
+  const id = title && makeId(title);
+  const target = <a className={style.target} id={id} />;
```
sethrcamp: `const Target = useCallback(() => <a className={style.target} id={id} />, [id]);`

src/components/subsection.jsx
```diff
@@ -10,60 +10,54 @@ import NoData from './noData.jsx';
...
+  const helpPopup = help && (
```
sethrcamp, line 16:
```
const HelpPopup = useCallback(() => (
    <>
        { help && (
            <span className="small ml-3">
                <HelpPopup id={`help-${title}`}>{help}</HelpPopup>
            </span>
        )}
    </>
), [title, help]);
```

src/components/subsection.jsx
```diff
@@ -10,60 +10,54 @@ import NoData from './noData.jsx';
...
-        {this.props.hardcoded && <span className="badge badge-danger">Hardcoded Example Data</span>}
-        {this.props.title && !this.props.hideTitle && renderTitle}
-        {typeof this.props.hasData !== 'undefined' && !this.props.hasData ? (
-          <NoData />
-        ) : (
-          <ErrorBoundary>{this.props.children}</ErrorBoundary>
-        )}
-      </div>
-    );
+  let renderTitle;
```
sethrcamp, line 22:
```
const TitleContent = useCallback(() => (
     <>
         {isMeta && <Target />}
         {title}
         <HelpPopup />
     </>
), [isMeta, title]);

const Title = () => {
    if(level === 1) return (<h4 className="text-break"><TitleContent /></h4>);
    if(level === 2) return (<h5><TitleContent /></h5>);
    return (<h3><TitleContent /></h3>);
};
```

src/components/subsection.jsx
```diff
@@ -10,60 +10,54 @@ import NoData from './noData.jsx';
...
+          {title}
+          {helpPopup}
+        </h3>
+      );
   }
-}
+
+  return (
+    <div className={style.subsection}>
+      {!isMeta && target}
```
sethrcamp: `{!isMeta && <Target />}`

src/components/subsection.jsx
```diff
@@ -10,60 +10,54 @@ import NoData from './noData.jsx';
...
+      {hardcoded && <span className="badge badge-danger">Hardcoded Example Data</span>}
+      {title && !hideTitle && renderTitle}
```
sethrcamp, line 45: `<Title />`

src/components/subsection.jsx
```diff
@@ -1,6 +1,6 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
```
sethrcamp, line 1: I made some suggestions in this file all based on the idea that rendering elements as variables is no longer preferred over rendering components.

I.e. this is no longer preferred:
```
const foo = ( <div>bar</div> );
return (
  <section>
    {foo}
  </section>
);
```
compared to this:
```
const Foo = () => ( <div>bar</div> );
return (
  <section>
    <Foo />
  </section>
);
```

What is trickier here is that all of these examples in this file rely on props, so useCallback is also used to improve efficiency. Alternatively, `Target`, `HelpPopup`, and `Title` could be moved outside of `Subsection` into their own stand alone components, and pass props to them:
```
<Target id={id} />
<HelpPopup help={help} title={title} />
<Title title={title} hideTitle={hideTitle} isMeta={isMeta} id={id} help={help} />
```
That version would not need the useCallbacks, but might be less intuitive to read.

src/containers/diseasePage/basicDiseaseInfo.jsx
```diff
@@ -14,126 +14,104 @@ import { smartAlphaSort } from '../../lib/utils';
...
-  }
-
-  renderSourceList(sources) {
-    return (
-      sources && (
-        <CommaSeparatedList>
-          {sources.map((source) => (
-            <ExternalLink href={source.url} key={`source-${source.source}`}>
-              {source.source}
+const renderTermList = (terms) =>
```
sethrcamp: All of these `render______` functions are esentially components, so they should use the component syntax. So for the names I would change to:
```
const renderTermList = (terms) => (           ->    const TermList = ({terms}) => (
const renderSourceList = (sources) => (       ->    const SourceList = ({sources}) => (
const renderDefinitionLinks = (links) => (    ->    const DefinitionLinks = ({links}) => (
const renderDefinition = (disease) => (       ->    const Definition = ({disease}) => (
```

src/containers/diseasePage/basicDiseaseInfo.jsx
```diff
@@ -14,126 +14,104 @@ import { smartAlphaSort } from '../../lib/utils';
...
-        <AttributeLabel>Sources of Associations</AttributeLabel>
-        <AttributeValue>{this.renderSourceList(disease.sourceReferenceLinkUrls)}</AttributeValue>
-      </AttributeList>
-    );
-  }
-}
+const BasicDiseaseInfo = ({ disease }) => (
+  <AttributeList>
+    <AttributeLabel>Definition</AttributeLabel>
+    <AttributeValue>{renderDefinition(disease.doTerm)}</AttributeValue>
```
sethrcamp: Similarly I would change the following:
```
{renderDefinition(disease.doTerm)}                     ->    <Definition disease={disease.doTerm} />
{renderTermList(disease.parents)}                      ->    <TermList terms={disease.parents} />
{renderTermList(disease.children)}                     ->    <TermList terms={disease.children} />
{renderSourceList(disease.sourceReferenceLinkUrls)}    ->    <SourceList sources={disease.sourceReferenceLinkUrls} />
```

src/containers/diseasePage/basicDiseaseInfo.jsx
```diff
@@ -14,126 +14,104 @@ import { smartAlphaSort } from '../../lib/utils';
...
+          </span>
+        );
+      })}
+    </CommaSeparatedList>
+  );
+
+const renderDefinition = (disease) =>
+  (disease.definition || (disease.definitionUrls && disease.definitionUrls.length > 0)) && (
+    <div>
+      {disease.definition} {renderDefinitionLinks(disease.definitionUrls)}
```
sethrcamp: Here I would change to:
```
{renderDefinitionLinks(disease.definitionUrls)} -> <DefinitionLinks links={disease.definitionUrls} />
```

src/containers/search/filterSelector/filterSelector.jsx
```diff
@@ -14,17 +14,16 @@ import CollapsibleFacet from './collapsibleFacet.jsx';
 import { selectActiveCategory, selectAggregations } from '../../../selectors/searchSelectors';
 import { selectPageLoading } from '../../../selectors/loadingSelector';
 
-class FilterSelectorComponent extends Component {
-  renderFilters() {
-    let aggs = this.props.aggregations;
-    if (aggs.length === 0) {
+const FilterSelectorComponent = ({ activeCategory, aggregations, isPending, queryParams }) => {
+  const renderFilters = () => {
```
sethrcamp, line 18: Similar to above comments, these two render functions should be converted to components

src/containers/search/resultsList.jsx
```diff
@@ -16,156 +16,128 @@ import SpeciesName from '../../components/SpeciesName.jsx';
...
-        </div>
-        <div className="col-sm-2">
-          <span className={style.resultCatLabel}>
-            <CategoryLabel category={d.category} />
-          </span>
-        </div>
-      </div>
-    );
-  }
+const renderHighlightedValues = (highlight, fields) => {
```
sethrcamp: Same comment as above about converting render functions to components

src/containers/search/resultsTable.jsx
```diff
@@ -10,112 +10,102 @@ import SpeciesName from '../../components/SpeciesName.jsx';
...
-        processedName = d;
-      }
-      return (
-        <th className={style.tableHeadCell} key={`srH.${d}`}>
-          {makeFieldDisplayName(processedName)}
-        </th>
-      );
-    });
-    return <tr>{nodes}</tr>;
+const renderTruncatedContent = (original) => {
```
sethrcamp, line 32: Same comment as above about converting render functions to components

src/reactApplication.jsx
```diff
@@ -22,28 +22,26 @@ const queryClient = new QueryClient({
...
-      <Provider store={store}>
-        <QueryClientProvider client={queryClient}>
-          <ReleaseContextProvider>
-            <Router>
-              {isBrowser ? (
-                <RouteListener onRouteChange={logPageView}>
-                  <LayoutWithRoutes />
-                </RouteListener>
-              ) : (
+const ReactApp = ({ router }) => {
```
sethrcamp: Couldn't this be: `const ReactApp = ({ Router = BrowserRouter }) => {`.
That would obviously require changes elsewhere too.

sethrcamp (changes requested):
I think the biggest comment(s) I have in general are to do with converting the "render_____" functions to components (see examples in comments).

The rest are minor and just things I happened to notice. As far as I can tell, all of these change should work identically, so the user should have no clue how much work you have put into these changes. But don't worry, the UI WG sees your efforts and thanks you!

src/components/subsection.jsx
Joe-Wdz: target is just JSX that you want to render, not a function that needs to be reused.

src/components/subsection.jsx
Joe-Wdz: Thanks for the suggestion! My understanding is that useCallback is intended for memoizing function references, typically when passing callbacks to memoized children or using them in dependency arrays.

In this case, target is just a JSX element that's rendered directly—it isn't being passed around as a callback or component. Using useCallback would wrap it in a function, which doesn't seem to provide a benefit here and makes the code a bit more complex.

Unless I'm missing a specific use case or performance concern, I'd prefer to keep the JSX element as it is.

src/components/subsection.jsx
Joe-Wdz: Thanks for the suggestion! I considered this approach, but I'm inclined to keep the current implementation.

My reasoning is that Title and TitleContent would only be local render helpers used once within Subsection. Introducing additional components (and useCallback for TitleContent) adds another level of indirection without providing a clear functional or performance benefit in this case.

I refactored it in a simpler way to reduce the duplicated heading markup.

src/components/subsection.jsx
Joe-Wdz: the same reason as above

src/components/subsection.jsx
Joe-Wdz: That makes sense,  I think it's better than using useCallbacks way above. because useCallbacks just call component later in same places. that didn't improve efficiency as much.

src/containers/search/resultsTable.jsx
Joe-Wdz: This one is just a function, there's no JSX in return.

src/reactApplication.jsx
Joe-Wdz: Good point, I refactored it as const ReactApp = ({ router: Router = BrowserRouter }) => {.
This keeps the component's public API unchanged, so existing usages don't need to be updated.

src/containers/search/filterSelector/filterSelector.jsx
Joe-Wdz: This is a render function declared inside the parent component and called only in the parent component, it seems no benefits to convert it to component outside the parent. but anyway I changed in term of suggests.

Joe-Wdz: it failed to pass claude review and claude suggests to remain as render function.

src/components/subsection.jsx
Joe-Wdz: the same reason as above

src/components/subsection.jsx
sethrcamp: When you wrap a functional component with `useCallback` it does still work as a component. So in my example, `<Target />` would still work properly. 

The reason for `useCallback` was to prevent `<Target />` from re-rendering when any of the other props besides "id" change (as an optimization). However, I have just learned that React has changed the method of how you implement this type of optimization. Instead of `useCallback`, `memo` (not to be confused with `useMemo`) is supposed to be used now.

Although the new `memo` function is cool, it seems overly complicated for such a small optimization in this case. 

I personally still prefer the component syntax over the element syntax (`<Target />` vs `{target}`) because I think it makes it clear than some kind of HTML is being rendered as apposed to plaintext. Since it is a matter of preference though, I am fine with it as is.

sethrcamp (approved):
I left an explainer of the useCallback stuff on one of the comments, but I think this looks good

src/components/subsection.jsx
Joe-Wdz: it makes sense, i ll change that.

PR #1816, opened 2026-07-15
feat(gene-page): switch sequence panel to tabix GFF3 adapter (KANBAN-…

scottcain:
…1452)

Tested locally and works correctly for 9.1.0.
generic-sequence-panel 2.0 replaces its default NCList feature-data adapter with tabix-indexed GFF3, ahead of the "All Genes" JBrowse migration. Bump the dependency, add a per-species jBrowseGffUrlTemplate alongside the existing VCF/FASTA URLs, and wire sequencePanelWrapper to pass gffurl instead of nclistbaseurl/urltemplate.


scottcain:
I'm marking this as a draft for the moment while I work on related changes.

scottcain:
Also note that at the moment, this doesn't work for zebrafish but they are working on updating their GFF.

oblodgett (approved):

PR #1820, opened 2026-07-17
SCRUM-6309 Fix crash when disease/phenotype annotations have no references

VarunReddy1111:
## Problem

On gene pages, checking **"Include negative annotations"** (and general loads for genes whose annotations have no references) crashed both the **Disease** and **Phenotype** sections with *"A problem displaying this section"*:

- **Disease:** `can't access property "sort", e is undefined`
- **Phenotype:** `can't access property "includes", t is undefined`

Reported against mouse **Shh** (`MGI:98297`), whose disease and phenotype annotations all return `pubmedPubModIDs: null`.

## Root cause

The References-column code assumed `pubModIds` was always an array. When it comes back `null`:

- `referencesCellCuration.jsx` called `getMultipleReferencesUrls(pubModIds)` **before** its null guard → `null.sort()` (Disease).
- `singleReferenceCellCuration.jsx` → `getSingleReferencePubModId` called `pubModIds.includes(...)` → `null.includes(...)` (Phenotype). This renders on load because reactstrap mounts the collapsed annotation-details dropdown into the DOM.

Backend responses were verified to be well-formed; this is purely a front-end null-handling bug.

## Fix

Guard all three reference helpers so a missing/empty reference list renders nothing instead of throwing:

- `referencesCellCuration.jsx` — return early when `pubModIds` is empty/not an array.
- `singleReferenceCellCuration.jsx` — `pubModIds?.includes(...)`.
- `utils.jsx` — guard `getMultipleReferencesUrls` / `getSingleReferenceUrl`, drop falsy entries, and **copy the array before sorting** to avoid mutating cached query data. Also hardens ~6 other call sites (interaction tables, reference page, variant summary).

## Verification

Reproduced the exact reported flow locally (dev server → stage API) on the Shh gene page:

- Page load → no error banners ✅
- Toggle "Include negative annotations" **on** → no errors ✅
- Toggle **off** (the original trigger) → no errors, tables render ✅


VarunReddy1111:
Closing this PR as this bug is fixed in SCRUM-6204 PR

PR #1824, opened 2026-07-22
fix(referencePage): render references with missing/null fields (KANBAN-1328)

jimflybase:
## KANBAN-1328 — Reference page not rendering for `AGRKB:101000000828456`

The reference is an `internal_process_reference` whose API `literatureSummary` omits many fields entirely (`cross_references`, `resource_title`, `volume`, `issue_name`, `page_range`). The rendering code accessed these unguarded, so the page threw before rendering — the reported blank page + console error `Uncaught TypeError: Cannot read properties of undefined (reading 'length')`.

### Changes

**`src/containers/referencePage/index.jsx`**
- Guard the MOD/external xref split loop with `cross_references || []` (this was the first crash / cause of the blank page).

**`src/containers/referencePage/ReferenceSummary.jsx`**
- Guard `date_published` before `.replace()`.
- Only run `Intl…formatToParts` when the date actually parses; year-only/invalid dates fall back to the raw value (avoids a `RangeError`).
- Default `volume`/`page_range` to `''` and omit empty citation segments so missing fields no longer render as literal `undefined`.
- Guard `category.replace()` → "Not Available" when absent.

### Verification
Ran the dev server against the stage API and loaded `/reference/AGRKB:101000000828456`:
- Page renders fully with **zero console errors**.
- Missing fields show "Not Available"; data tables degrade to "No data available".
- Publication Source renders cleanly with no leading `undefined.`

Context: per Chris Grove, this reference will be linked from all predicted disease annotations (via_orthology) on disease pages in 9.1.0, so it needs to render.

### DIFF
--- a/src/containers/referencePage/ReferenceSummary.jsx
+++ b/src/containers/referencePage/ReferenceSummary.jsx
@@ -29,8 +29,8 @@ const CommaSeparatedSourceList = ({ sources }) => {
 const PubSourceLink = ({ ref }) => {
   const publisher = ref.resource_title;
   // this should not need to be done here; talk to Blue Team about fixing these date formats (e.g. "1999.7.30")
-  // date_published may be missing entirely (e.g. internal_process_reference)
-  const date_pub_fixed = ref.date_published ? ref.date_published.replace(/\b(\d)\b/g, '0$1').replace(/\.+/g, '-') : '';
+  // date_published may be missing entirely (e.g. internal_process_reference); ?. leaves it undefined (falsy) then
+  const date_pub_fixed = ref.date_published?.replace(/\b(\d)\b/g, '0$1').replace(/\.+/g, '-');
   if (!publisher && !date_pub_fixed) return <NoData>Not Available</NoData>;
   // date_published may be a full date ("1999-07-30") or year-only ("2018"); only reformat when it parses,
   // otherwise show the raw value (a year-only string produces an Invalid Date that formatToParts throws on)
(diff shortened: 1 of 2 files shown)


src/containers/referencePage/ReferenceSummary.jsx
```diff
@@ -29,22 +29,29 @@ const CommaSeparatedSourceList = ({ sources }) => {
 const PubSourceLink = ({ ref }) => {
   const publisher = ref.resource_title;
   // this should not need to be done here; talk to Blue Team about fixing these date formats (e.g. "1999.7.30")
-  const date_pub_fixed = ref.date_published.replace(/\b(\d)\b/g, '0$1').replace(/\.+/g, '-');
+  // date_published may be missing entirely (e.g. internal_process_reference)
+  const date_pub_fixed = ref.date_published ? ref.date_published.replace(/\b(\d)\b/g, '0$1').replace(/\.+/g, '-') : '';
```
sethrcamp: can be simplified to:
```
const date_pub_fixed = ref.date_published?.replace(/\b(\d)\b/g, '0$1').replace(/\.+/g, '-') ;
```
which will make `date_pub_fixed` undefined if `ref.date_published` is missing. Since `undefined` is also falsely, your logic in the next line should be okay still

sethrcamp (approved):
Left a minor improvement, but looks good to me

src/containers/referencePage/ReferenceSummary.jsx
jimflybase: Good call — applied in b3c35a0c. Switched to `ref.date_published?.replace(...)`; the `undefined` result stays falsy so the downstream guards behave the same. Thanks for the review!

PR #1833, opened 2026-07-28
simple fix to correct yeast chromosome name munging

scottcain:


scottcain:
I do this exact same munging in a few different places which I will probably pull out into a helper function either for this release or the next one, but in the meantime, this should be merged so that orthology/synteny links out to JBrowse work.

cmpich (approved):

PR #1834, opened 2026-07-28
KANBAN-1322 ontology browser: reliable deep-link scroll, no-jump tree clicks

jdepons:
## Summary
Follow-up polish for the ontology browser, all scoped to `src/containers/ontologyBrowser/`.

- **Batch fetcher URL length.** `/api/disease/batch` requests now cap at 80 curies per URL. Stage's WAF was returning 403 on batches large enough to push the URL past ~1800 chars, which dropped the term docs needed to expand deeper ancestor chains.
- **No-jump in-tree clicks.** `fromTree` is now tracked in a ref that updates synchronously inside the click handler, so it is guaranteed correct on the very first render after the click (no batching race with router state updates). The anchor scroll additionally skips when the focused row is already visible inside its scroll ancestor, so an in-tree click never moves the pane away from the user's cursor.
- **Steadier deep-link load.** Tree stays hidden behind a DNA loading indicator until the focused row mounts, then reveals in one paint instead of jumping while ancestor batches trickle in. A 3s fallback reveals the tree if the curie is invalid / has no ancestors.
- **Badge order.** Annotation-available badges reordered to **Genes / Alleles / Models**.

## Test plan
- [ ] Load a deep term URL (e.g. `/ontology/disease/DOID:0111251`) and confirm the tree reveals with the term centered and highlighted.
- [ ] Load an invalid curie (e.g. `/ontology/disease/DOID:0060906`) and confirm the tree reveals to the root after ~3s.
- [ ] Expand many ancestors, scroll the tree pane, click a deep term — the row should stay put, no upward jump.
- [ ] Confirm annotation badges render in Genes → Alleles → Models order in the legend and per-row.

### DIFF
--- a/src/containers/ontologyBrowser/OntologyTree.jsx
+++ b/src/containers/ontologyBrowser/OntologyTree.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faChevronRight, faChevronDown } from '@fortawesome/free-solid-svg-icons';
-import { Link } from 'react-router-dom';
 import CountBadge from './CountBadge.jsx';
 import { ANNOTATION_TYPES } from './annotationTypes.js';
 import { useDiseaseTerm } from './useDiseaseTerms.jsx';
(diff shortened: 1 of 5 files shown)


jimflybase (approved):
I've run the suggested tests, and they were passed.

PR #1836, opened 2026-07-28
KANBAN-864: add local AGRKB Reference column to report tables

jimflybase:
## Dependency satisfied — agr_java_software#1649 is merged

**Requires: alliance-genome/agr_java_software#1649 — merged into `stage` on 2026-07-28 as `6948fe268`.**

The new **Reference** column is `filterable` with `filterName: 'referenceCitation'`, which sends a `filter.referenceCitation` query param. That param was added by #1649, so this PR no longer has an unmet dependency and the earlier "do not merge alone" warning no longer applies.

What this means for review:

- **Filtering works against any API at or past `stage` `6948fe268`.** Confirmed on `stage`: `references.shortCitation` is registered in `DiseaseController` (×4), `AlleleController` (×2) and `GeneController` (×1), covering all 7 endpoint registrations plus their `/download` variants.
- **An Amplify preview only filters once its API environment has picked up that `stage` commit.** Until then the column still renders and links correctly, but the row count will not change — and note the de-emphasis *is* applied client-side from the filter term regardless, so the table looks responsive (matches stand out against dimmed neighbours) while pagination stays put. That is an artifact of the API environment lagging, not a bug in this PR. Verified behaviour on the preview before the merge: `GET /api/disease/DOID:14330/genes?filter.referenceCitation=Tang` returned 200 with the count unchanged at 3,312, 9 links dimmed and the 2 matching links full-strength.
- Against a local API carrying the change, the same filter narrows 3,312 → 2.

## Summary

Adds a local AGRKB **Reference** column (citation text linking to the in-app `/reference/{curie}` page) *alongside* the existing external PMID column, whose header is renamed `References` → **Reference ID**. The PMID column is deliberately retained — some users may want PMID filterability — so the end state is both columns side by side.

When the Reference filter is active, non-matching citations within a cell are de-emphasized (`opacity: 0.5` + `saturate(0.55)`) rather than hidden, so it stays visible which reference in a multi-reference cell matched.

## Surfaces covered

| surface | commit |
| --- | --- |
| `DiseaseToGeneTable`, `DiseaseToAlleleTable`, `DiseaseToModelTable`, `AlleleToDiseaseTable`, `genePage/phenotypeTable` (gene + allele pages) | `f8e94cbf`, `c2a34dbe` |
| Gene-page disease ribbon table (`components/disease/diseaseAnnotationTable.jsx`) | `5c143b74` |
| Annotation-detail popups (entities, entities-gene-page, phenotype) | `84abb408` |

Note that `genePage/phenotypeTable.jsx` is shared by the gene and allele pages, so 5 changed table components cover 6 API endpoints.

## Implementation notes for review

- **`referenceList.jsx`** renders `references[]` (array) with a `filterTerm` prop driving the de-emphasis; containers thread `tableProps.tableState?.filters?.referenceCitation?.filterVal` into the formatter.
- **`singleReferenceLinkCuration.jsx`** (new, for popups) renders a single `entity.evidenceItem`. It only emits a link for `AGRKB:` curies and falls back to plain text otherwise. This guard is load-bearing: on `/api/gene/HGNC:11998/phenotypes`, annotation evidence is largely `ExternalDatabaseReference` items — 98 `ORPHA:` and 37 `MIM:` with no `shortCitation`, versus only 14 AGRKB `Reference` items. Without the guard those rows would link to a nonexistent `/reference/ORPHA:96253`. Consequence: the popup Reference cell is blank for those rows. Disease annotations are unaffected (547/547 sampled were AGRKB).
- **Reference pages omit the new column** via `REFERENCE_PAGE_DETAILS_COLUMNS`, since every annotation there already belongs to the page's reference and the citation would repeat on every row.
- `5c143b74` also flips `let columns` → `const columns` in `diseaseAnnotationTable.jsx`, clearing the one pre-existing eslint error in that file.

## Verification

Verified in a browser against a local `agr_api` carrying #1649. Filtering the gene-page disease ribbon by `Bally` narrows 85 rows → 2, with the local `/reference/AGRKB:101000000462421` link in **Reference** and external `PMID:24836762` in **Reference ID**. Popups verified on the gene ribbon, the disease-page Genes table, and the gene-page Phenotypes table; the reference-page popup correctly ends at `Source | Reference ID`.

Prettier and eslint are clean on all changed files. (The annotation popups carry 9 pre-existing eslint errors — `var expCondition`, unused `mainRowCurie`, and a missing `jsx-a11y/anchor-is-valid` rule definition — which predate this branch and are left alone.)

## Out of scope

Three tables cannot get this column: both gene interaction detail tables and the expression ribbon. Their documents expose no AGRKB reference at all — only `crossReferences` (DOI / PMCID / `biogrid:` / MOD ids) — so this is a data-availability blocker upstream, not UI work.


jimflybase:
Thanks — taking 1, declining 2, leaving 3 to the ticket owner. Details below since 2 turned out not to reproduce.

## 1. `filterTerm` shape — fixed in 064ce930

Confirmed the mechanism: `DataTable.jsx:145-165` selects the renderer from the API response, and `DropdownCheckboxFilter`'s `onChange` emits an array, so the shape is decided outside this component. A throw from a cell formatter takes the table down, not the cell, so the formatter should be total regardless of how likely the array is.

Worth recording how far away the trigger actually is, though: the auto-detect path calls `getDistinctFieldValue(supplementalData, 'filter.' + filterField)`, i.e. it looks up `distinctFieldValues['filter.referenceCitation']`, and the API doesn't use that key convention at all — `/api/disease/DOID:14330/genes` returns bare keys (`associationType`, `diseaseQualifiers`, `species`). So today that lookup returns `[]` for *every* column, and the columns that do get checkbox filters pass an explicit array as `column.filterable`. Firing this would take both a new aggregation and a `filter.`-prefixed key.

Fixed anyway, and taking the second suggestion rather than the join, since concatenating selections would produce an unmatchable string:

```js
const terms = (Array.isArray(filterTerm) ? filterTerm : [filterTerm])
  .map((value) => String(value ?? '').trim().toLowerCase())
  .filter(Boolean);
// ...
const dim = terms.length > 0 && !terms.some((term) => citation.includes(term));
```

Checked against every shape — `'Bally'`, `['Bally']`, `['Tang','Bally']`, `['Tang','Ko']`, `undefined`, `null`, `''`, `[]`, `['']`, `[null,'Bally']`, `['  Bally  ']` — no throws, and multi-select now behaves as OR. The previous implementation threw `TypeError: (filterTerm || "").trim is not a function` on any array. Re-verified in the browser against a local API: filtering the disease-page Genes table by `Tang` narrows 3,312 → 2, with `Tang G` and `Tang FL` at full strength and the non-matching `Chiu CC (2020)` dimmed.

## 2. De-duplication — not reproducible; declining

The three supporting points don't hold up:

- **The cited `DiseaseToGeneTable.jsx` dedupe does not exist.** There is no `new Set` and no `removeDuplicates` anywhere in that file. Its via-orthology branch flatMaps `references` to build PMID xrefs out of `pubModID`; it never dedupes.
- **`removeDuplicates` in `referencesCellCuration.jsx` is real but operates on a different array** — it keys on `referencedCurie` over the `pubmedPublications` xrefs, not over `references[]`.
- **No duplicates in the data.** 500 rows sampled across every endpoint this column appears on — `disease/DOID:14330` genes/alleles/models, `gene/HGNC:11998/phenotypes`, `allele/RGD:7241046` phenotypes and diseases, and the `POST /api/disease` ribbon — including 40 multi-reference rows: zero rows where `references[]` repeats a curie.

Happy to revisit if someone can point at a row that actually repeats a curie, since the fix is cheap — but as it stands the premise is the reverse of what's in the code.

## 3. `.referenceDim` contrast — real, and a product decision

The measurement is fair: `opacity: 0.5` + `filter: saturate(0.55)` does put de-emphasized citations under AA, and match/no-match is carried by lightness alone. Worth adding that the dimmed items remain live links, which sharpens the point.

De-emphasising non-matches (rather than highlighting matches) is the UX that was explicitly chosen for this ticket, so I'm not changing it unilaterally — flagging it for @jimflybase to decide whether to soften the dim or pair it with a non-color cue. Not a merge blocker either way.

jimflybase:
**On point 3:**
I will await tester feedback and adjust as needed.

jimflybase:
Merged `stage` into this branch (0a7880d3) — no conflicts, and the merged tree is green locally (43 suites / 192 tests).

### Intentional overlap with #1827 — please read before flagging it

`stage` picked up #1827 / SCRUM-6204 (@adamgibs), which changed the via-orthology branch of the `pubmedPublications` column to render AGRKB curies as internal `/reference/{curie}` links — because via-orthology rows have no PMID (`pubmedPublications` comes back as `[{ curie: null }]`).

This PR adds a `Reference` column that renders `shortCitation`, also linking to `/reference/{curie}`. The two changes auto-merge cleanly, but on via-orthology rows the result is two adjacent columns pointing at the **same href**:

| Gene | Species | Association | Reference | Reference ID |
|---|---|---|---|---|
| Ddr1 (Rno) | Homo sapiens | is implicated via orthology | The Alliance of Genome … | AGRKB:101000000828456 |
| Ep300 (Rno) | Homo sapiens | is implicated via orthology | The Alliance of Genome … | AGRKB:101000000828456 |
| BRD4 (Hsa) | Rattus norvegicus | is implicated via orthology | The Alliance of Genome … | AGRKB:101000000828456 |

Verified on `/disease/DOID:0050156` → Associated Genes with the Association column filtered to *is implicated via orthology* — all 101 rows behave this way, and it will apply to every via-orthology row on every disease.

**We're deliberately shipping both for 9.1.0** rather than resolving it now, to get curator feedback on which column they actually want before removing either. Options if the feedback says to collapse them:

- leave `Reference ID` empty on via-orthology rows (there genuinely is no PMID) — a small change to the `getIsViaOrthology` branch in `DiseaseToGeneTable.jsx`, but it would revert #1827's behaviour, so it needs @adamgibs' agreement;
- show the curie only when the `Reference` column is empty.

### Known data issue visible in the new column

The `shortCitation` on the orthology reference renders as `The Alliance of Genome Resources Curators (2018)   ():` — empty journal/volume parens, and the title is dropped even though the record has one. That's reference-data generation rather than a UI bug; filed as KANBAN-1465. It only became visible because this PR surfaces citations.

jdepons (approved):

PR #1837, opened 2026-07-28
KANBAN-678 reference page: simplify to Summary + Abstract until Reference sections are ready

jdepons:
## Summary
Simplify the reference page on stage down to **Summary + Abstract** while the four Reference-array table sections stay dormant on disk for local iteration.

### What ships to stage
- Reference page renders only the Summary and Abstract subsections (see `src/containers/referencePage/index.jsx`)
- The five older annotation-driven sections are **removed** from the codebase: Phenotypes, Disease Associations, Expression, Molecular Interactions, Genetic Interactions (and their table components)
- The four new Reference-array sections — Genes, Alleles/Variants, Transgenic Alleles, Models — have their `<Subsection>` blocks, `sections` array entries, and `useSectionCounts` call **commented out** with restoration notes inline. No count queries fire while the sections are hidden.

### What stays on disk (not rendered)
- `tables/ReferenceGeneTable.jsx`
- `tables/ReferenceAlleleTable.jsx`
- `tables/ReferenceTransgenicAlleleTable.jsx`
- `tables/ReferenceModelTable.jsx`
- `createReferenceTable.jsx`
- `getSingleReferenceUrl` shim in `src/components/dataTable/utils.jsx`

These come back to the page by uncommenting three blocks in `index.jsx` — the sections array entries, the `useSectionCounts` call, and the four `<Subsection>` renders.

### Depends on
- Backend PR alliance-genome/agr_java_software#1650 for the `/api/reference/{id}/{genes,alleles,transgenic-alleles,models}` endpoints that the on-disk tables will call once they're restored. Since the sections are hidden on stage, no dependency is enforced at merge time.

## Test plan
- [ ] `/reference/AGRKB:101000000133151` renders with Summary + Abstract only, no PageNav badges for the removed sections
- [ ] No 404s or 500s in the network tab from the removed endpoints (`disease-annotations`, `phenotype-annotations`, `expression-annotations`, `molecular-interactions`, `genetic-interactions`)
- [ ] No count requests fire for `genes`, `alleles`, `transgenic-alleles`, or `models`
- [ ] Uncommenting the three marked blocks in `index.jsx` restores the four sections locally


jimflybase (changes requested):
Requesting changes to hold the merge: items **1**, **2** and **4** from the automated review above still stand on the current head `91dc597a`, and `claude-review` is `skipped` on that commit, so nothing has re-checked them since the `origin/stage` merge two commits later. Not re-litigating those items — just adding what has changed since Jul 28, because the urgency is now higher than when they were written.

**1. The crash is confirmed against live data, not just possible.** Item 1 said "any such reference." There is one, and it is not obscure:

```
GET https://stage.alliancegenome.org/api/reference/AGRKB:101000000828456
  category: Internal_Process_Reference
  has cross_references key: False
  resource_title: None   volume: None   page_range: None
```

The key is absent, so `index.jsx:134` throws `TypeError: Cannot read properties of undefined (reading 'length')`. The same record also has `resource_title` / `volume` / `page_range` all null, so it trips item 4's `undefined. 2018; undefined :undefined` string as well.

**2. It blanks the entire app, not just the reference page.** The throw is in the `ReferencePage` body above `<DataPage>`, and there is no ancestor ErrorBoundary — `routes.jsx:46` renders `<ReferencePage />` bare, and the boundaries in `subsection.jsx` / `dataPage.jsx` all sit below the throw. So there is no section-level fallback; React unmounts to the root.

**3. #1836 landed today and puts that exact record one click away.** `AGRKB:101000000828456` is the reference on every via-orthology disease annotation, and #1836 turns AGRKB curies into internal `/reference/{curie}` links across the disease tables. When the Jul 28 review was written this crash was reachable mainly by direct URL; on current `stage` it is reachable by clicking a reference link on a disease page.

**4. Two of the earlier items are now moot.** Commit `7dff0d16` commented out the four `<Subsection>` blocks, so items **5** (`AlleleCell` guard / `alleleDocument` shape) and **6** (paging-dependent variant columns) are now in code that does not render — worth keeping for whenever the sections are restored, but not merge blockers. Item **3**'s `getSingleReferenceUrl` shim *is* still live: `ReferenceSummary.jsx` and `SourceList` both call it, so it is the URL behind the Cross References, MOD Identifier and DOI citation links.

**On root cause** — one correction to the earlier review's guess that the branch was cut before these fixes landed. The merge-base with stage, `f14ee5c9`, already had the guards; they are absent in `b03dab9c`, this branch's first commit. So that commit rewrote the block into the older unguarded shape rather than merely predating the fix, which is also why the later `origin/stage` merge did not restore it.

**Suggested fix:** take `stage`'s versions of the `cross_references` loop in `index.jsx` and of `PubSourceLink` + the Alliance Publication Type cell in `ReferenceSummary.jsx`, rather than re-deriving the guards. That clears 1, 2 and 4 together and keeps the shim question (item 3) separable from this PR's actual subject.

jimflybase (approved):
Approving — this supersedes my `CHANGES_REQUESTED` from Jul 30. Commit `61032341` clears the blocker, and I verified it rather than taking the commit message for it.

**The crash is fixed.** `AGRKB:101000000828456` still has no `cross_references` key on stage (re-fetched today), and `for (const entry of ref.cross_references || [])` no longer throws on it. Traced the same record through the new `PubSourceLink`: `date_published: '2018'` is year-only, the `date_pub_fixed && !isNaN(...)` guard skips the `Intl` reformat instead of throwing, and the empty-segment filter yields a citation of `2018` rather than the old `undefined. 2018; undefined :undefined`. Items 1, 2 and 4 are genuinely resolved, and taking stage's versions rather than re-deriving the guards was the right call.

**Verified against a real merge into current `stage`, not just the diff:**

- Merges clean with no conflicts.
- `vite build` succeeds — this matters because six table files are deleted here and `tsc` doesn't resolve `.jsx` imports, so a dangling import would only surface at bundle time. None.
- 43/43 suites, 192/192 tests pass.
- `tsc --noEmit` clean.

**One thing I checked specifically, since it looked dangerous and isn't:** dropping `getSingleReferenceUrl` reads like it would orphan `getMultipleReferencesUrls`, which still calls a function of that name on `stage`. It doesn't. The shim was added and removed entirely within this branch, so `utils.jsx` has a net-zero diff and isn't in the changed-file list; stage's real `getSingleReferenceUrl` (the `getResourceUrl` one, from the `main`→`stage` mergeback) is a different function that survives the merge intact. Confirmed present and called in the merged tree.

Non-blocking, for awareness rather than action here:

1. **This branch is `behind`** and needs the update-branch step before it can merge. My trial merge says that will be uneventful.
2. **The `ReferenceSummary.jsx` cleanup here is byte-identical to a commit in my open #1843** — same seven unused imports/`modMap` removals, same `/^doi\:/i` → `/^doi:/i` fix, same Prettier reflow. We did the same cleanup independently. They merge cleanly in either order and the duplication is harmless; whichever lands second just contributes nothing to that file. I also merged #1843 and #1842 on top of this in the trial and all three coexist without conflict, with the KANBAN-1153 field order and the KANBAN-1151 sidebar entry both intact.
3. **Items 5 and 6 from the Jul 28 automated review are still unaddressed** (the `AlleleCell` guard / `alleleDocument` shape, and the paging-dependent variant columns). They sit inside the four `<Subsection>` blocks that this PR comments out, so they're unreachable at merge time and correctly not blockers — but they'll be live again the moment those sections are restored, so they're worth carrying into whatever ticket does that.

PR #1860, opened 2026-08-17
fix(crossReferenceList): sort a copy instead of mutating the prop (KANBAN-1325)

jimflybase:
## Summary

`crossReferenceList.jsx` called `crossReferences.sort()` directly on its prop. `Array.prototype.sort` mutates in place, so rendering reordered the caller's array — a React immutability violation. This is the same bug fixed for the curation variant (`CrossReferenceListCuration.jsx`) in 08461f05; this is the pre-existing non-curation copy.

Fix: sort a copy and map over it.

## Scope / impact

No live user-visible impact today, so this is hygiene rather than a user-facing fix:

- The disease page (`basicDiseaseInfo.jsx`, the only live caller) passes a freshly built array from `transformCrossReferenceLinkUrls`, so the mutated array is a throwaway.
- The other caller, `basicGeneInfo.jsx`, passes cached query data — but that file is imported nowhere and is dead code. The live gene-page cross references render through `GeneSummary` -> `CrossReferenceListCuration`, already fixed.

The value is that the next caller to pass cached data doesn't silently corrupt the query cache.

## Tests

New `src/components/crossReferenceList.test.jsx` covers the ordering contract, prop immutability, and the `sort={false}` passthrough. Companion e2e coverage of the disease-page ordering is proposed separately in `agr_release_testing`.

KANBAN-1325

### DIFF
--- a/src/components/crossReferenceList.jsx
+++ b/src/components/crossReferenceList.jsx
@@ -22,13 +22,12 @@ const CrossReferenceList = ({ collapsible = true, crossReferences, sort = true }
 
   const size = collapsible ? undefined : crossReferences.length;
 
-  if (sort) {
-    crossReferences.sort(byWithLinkThenName);
-  }
+  // Copy before sorting — sort mutates in place and crossReferences is a prop.
+  const orderedReferences = sort ? [...crossReferences].sort(byWithLinkThenName) : crossReferences;
 
   return (
     <CollapsibleList collapsedSize={size}>
-      {crossReferences.map((ref) => (
+      {orderedReferences.map((ref) => (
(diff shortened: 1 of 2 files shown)


christabone (approved):
Reviewed the component contract, both callers, the existing CrossReferenceListCuration implementation, and the new regression tests. Copying before sort removes the React prop mutation while preserving linked-first/name ordering and the sort={false} passthrough; the approach matches the established curation variant and retains compatibility without requiring Array.prototype.toSorted. The focused Jest suite passes (3/3), the production build succeeds, the diff is clean, and all required GitHub checks pass. No material findings.

PR #1862, opened 2026-08-18
feat(interactions): add Alliance Reference column to interactions tables (KANBAN-1485)

jimflybase:
Adds the Alliance Reference (short citation) column to the two gene-page interactions tables. Jira: [KANBAN-1485](https://agr-jira.atlassian.net/browse/KANBAN-1485)

## Summary

Both tables already had a column headed "Reference", but it rendered `evidence[0].referenceID` as an outbound PubMed link — functionally the "Reference ID" column of the other tables under the wrong label, which is likely why the gap was not obvious. So:

- Renamed the existing column to **Reference ID** (unchanged behavior, still `filterName: 'reference'`).
- Added a **Reference** column before it, rendering `evidence[]` through `ReferenceList` — the same treatment KANBAN-864 gave the Phase 1 tables. Links to `/reference/<curie>`.

No API work needed; the endpoints already return `curie` and `shortCitation` in `evidence[]`.

### Why the new column uses a dummy dataField

`react-bootstrap-table-next` keys header and filter cells by `dataField` (`header.js:84`, `filters.js:47`). Two columns both pointing at `evidence` would collide on that key and could misplace the filter element, which belongs only to Reference ID. The new column is therefore `isDummyField: true` with its own `dataField`, reading evidence off the row. The existing column is left completely untouched so its filter wiring still resolves.

### Citation filter — now enabled

The new column carries a free-text filter (`filter.referenceCitation`), the same
widget the Reference ID column has.

This shipped unfiltered at first, because the server side did not exist yet and these
endpoints ignore unknown filter params *silently* — a filter box that returned the full
result set would have looked functional. That is resolved:
[KANBAN-1487](https://agr-jira.atlassian.net/browse/KANBAN-1487) merged as
[agr_java_software#1662](https://github.com/alliance-genome/agr_java_software/pull/1662)
on Aug 18, and its merge-triggered reindex has landed, so `evidence.shortCitation` is
genuinely indexed on stage. Verified against the stage API before enabling the UI filter:
`filter.referenceCitation=Tanaka` on `w`'s genetic interactions returns 2 of 159, and a
nonsense string returns 0 — i.e. neither silently ignored nor zero-matching.

Note the `test` API deploy predates that merge and still ignores the param, so review the
filter on stage or on the branch preview, not on test.

Matching is case-insensitive substring and multi-token, so `Nat Commun`, `2024`, and
`giot l (2003)` all work.

## Verification

Ran the local UI against the **test** API on `/gene/FB:FBgn0003996` — stage's API was returning 502 on every endpoint when this was opened — both tables:

- Reference renders e.g. `Tanaka T (2024) Nat Commun 15(1):6993` → `/reference/AGRKB:101000001041200`; Reference ID still renders `PMID:39143098`.
- Column order is Reference then Reference ID.
- Applying the Reference ID filter still sends `filter.reference=PMID%3A39143098` and narrows to 2 rows.
- Filter, re-verified on `/gene/HGNC:11998` against the **stage** API after KANBAN-1487 landed: molecular 5,594 → **2** rows on `Desantis A (2015)`, genetic 385 → **105** on `Xie L`, `Clear` restores the full set.
- The two filters are independent — with both applied the request carries `filter.reference=PMID%3A23284306&filter.referenceCitation=Xie%20L`, so the dummy `dataField` avoids the header/filter key collision it was introduced to avoid.
- Console key-warning count is identical to clean `stage` (2, from `Body2`/`Cell2`, pre-existing and from another table) — the dummy field adds none.
- `tsc --noEmit` clean; Jest 44 suites / 195 tests pass; Prettier clean. The 2 ESLint errors in these files are pre-existing on `stage` (an `eslint-disable` for a rule this config does not define).

Rendering more than the first evidence entry in Reference ID is out of scope — [KANBAN-1486](https://agr-jira.atlassian.net/browse/KANBAN-1486). Note the new Reference column already renders the whole array.

[KANBAN-1485]: https://agr-jira.atlassian.net/browse/KANBAN-1485?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KANBAN-1487]: https://agr-jira.atlassian.net/browse/KANBAN-1487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
+++ b/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
@@ -191,6 +191,8 @@ const GeneGeneticInteractionDetailTable = ({ focusGeneId, focusGeneDisplayName }
           width: '180px',
         },
         formatter: (_, row) => <ReferenceList refs={row.geneGeneticInteraction?.evidence} />,
+        filterable: true,
+        filterName: 'referenceCitation',
       },
       {
         dataField: 'geneGeneticInteraction.evidence',
--- a/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
+++ b/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
@@ -1,6 +1,6 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
-import { DataTable, GeneCellCuration, SpeciesCell } from '../dataTable';
+import { DataTable, GeneCellCuration, ReferenceList, SpeciesCell } from '../dataTable';
 import { getIdentifier } from '../dataTable/utils.jsx';
 import { buildUrlFromTemplate } from '../../lib/utils.js';
 import SpeciesName from '../SpeciesName.jsx';
@@ -183,8 +183,18 @@ const GeneGeneticInteractionDetailTable = ({ focusGeneId, focusGeneDisplayName }
         filterName: 'source',
       },
       {
-        dataField: 'geneGeneticInteraction.evidence',
+        // dummy field: the Reference ID column below owns the real evidence dataField
+        dataField: 'referenceCitation',
+        isDummyField: true,
         text: 'Reference',
+        headerStyle: {
+          width: '180px',
+        },
+        formatter: (_, row) => <ReferenceList refs={row.geneGeneticInteraction?.evidence} />,
+      },
+      {
+        dataField: 'geneGeneticInteraction.evidence',
+        text: 'Reference ID',
         headerStyle: {
           width: '150px',
         },
--- a/src/components/interaction/genePhysicalInteractionDetailTable.jsx
+++ b/src/components/interaction/genePhysicalInteractionDetailTable.jsx
@@ -120,6 +120,8 @@ const GenePhysicalInteractionDetailTable = ({ focusGeneDisplayName, focusGeneId 
       headerStyle: { width: '11em' },
       headerClasses: style.columnHeaderGroup3,
       classes: style.columnGroup3,
+      filterable: true,
+      filterName: 'referenceCitation',
     },
     {
       dataField: 'geneMolecularInteraction.evidence',
--- a/src/components/interaction/genePhysicalInteractionDetailTable.jsx
+++ b/src/components/interaction/genePhysicalInteractionDetailTable.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { DataTable, GeneCellCuration, SpeciesCell } from '../dataTable';
+import { DataTable, GeneCellCuration, ReferenceList, SpeciesCell } from '../dataTable';
 import SpeciesName from '../SpeciesName.jsx';
 import { getIdentifier } from '../dataTable/utils.jsx';
 import { buildUrlFromTemplate } from '../../lib/utils.js';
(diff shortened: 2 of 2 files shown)


jimflybase:
Update since this was opened — nothing in the diff has changed, but three things around it have.

**1. Re-verified against stage.** The original verification used the `test` API because stage was returning 502 on every endpoint at the time. Stage is healthy again, so I re-ran it there on `/gene/HGNC:11998` (TP53). Both tables render correctly:

| Table | Reference | Reference ID |
| --- | --- | --- |
| Molecular Interactions | `Liu J (2020) Nat Cell Biol 22(9):1056-1063` → `/reference/AGRKB:101000001018078` | `PMID:32807901` |
| Genetic Interactions | `Xie L (2012) PLoS Genet 8(12):e1003151` → `/reference/AGRKB:101000000456983` | `PMID:23284306` |

Reference sits immediately before Reference ID in both (column indices 6/7 and 9/10). Stage's API returns `curie` and `shortCitation` in `evidence[]` on both endpoints, as expected.

**2. The filter is blocked on more than KANBAN-1487 — I've corrected the PR body.** It previously said the citation filter simply waits on KANBAN-1487. That is now open as [agr_java_software#1662](https://github.com/alliance-genome/agr_java_software/pull/1662), and building it surfaced a second dependency: both interaction objects are mapped `dynamic: false` in Elasticsearch, with `evidence.referenceID` as the only indexed subfield under `evidence`, so `evidence.shortCitation` was present in `_source` but never indexed. #1662 adds it to `Mapping.java`, but a mapping change only applies to a **newly created** index — so until `site_index` is rebuilt, a citation filter here would return *zero* rows rather than unfiltered ones. Worth knowing before anyone reads "1487 merged" as "the UI filter is unblocked."

> **Corrected after posting:** @oblodgett confirms every merge triggers a new reindex, so the mapping goes effective when #1662 lands rather than needing a separately scheduled rebuild. The interim behavior above still holds — an unmapped field yields 0 rows with a 200 — but no reindex needs chasing.

**3. The red `claude-review` check is not about this code.** That job failed on its own after one turn (`is_error: true`, 5.6s, no message) and posted no review, so there are no findings to address. `prettier-check`, `test-install-build`, and GitGuardian are all green. Happy to re-trigger it if that's the convention here.

Still needs a human review — this has had none yet, bot or otherwise.

cerivs (commented):
@jimflybase  I think it looks ok but I don't have  a site to do check how the site looks with this code.
Is there a UI person who can do a true code review or should I just approve?

jimflybase:
@cerivs try this Amplify:  https://feature-kanban-1485-interactions-reference-column.d1mrdtdifcvuyq.amplifyapp.com/

jimflybase:
@cerivs I'm wiring up the filtering on those new columns now; the reindex is finished.  These changes should show up in that Amplify instance when I commit.

Done; it should be there as soon as this instance rebuilds.

jimflybase:
### Where to review this

Branch preview (rebuilt with the citation filter, backend is stage-like so the filter is live):

**https://feature-kanban-1485-interactions-reference-column.d1mrdtdifcvuyq.amplifyapp.com/**

Suggested pages — both have the Molecular *and* Genetic Interactions tables populated:

| Page | Molecular | Genetic |
| --- | --- | --- |
| [`/gene/HGNC:11998`](https://feature-kanban-1485-interactions-reference-column.d1mrdtdifcvuyq.amplifyapp.com/gene/HGNC:11998) (human TP53) | 5,594 | 385 |
| [`/gene/FB:FBgn0003996`](https://feature-kanban-1485-interactions-reference-column.d1mrdtdifcvuyq.amplifyapp.com/gene/FB:FBgn0003996) (fly `w`) | 3 | 159 |

What to check in each table:

- **Reference** shows the Alliance short citation, linking to `/reference/AGRKB:...`; **Reference ID** still shows the PMID (or MOD ref) linking out.
- Both columns filter independently. On TP53, molecular `Desantis A (2015)` → 2 rows; genetic `Xie L` → 105 of 385. Matching is case-insensitive substring, so `Nat Commun` or `2024` work too.

Two things that are *data*, not this PR:

- Zebrafish and mouse genes have no genetic-interaction data at all (e.g. ZFIN tp53 `ZDB-GENE-990415-270`), so an empty Genetic Interactions section there is expected.
- Some non-PubMed MOD references have ragged citations — fly `w` shows `Mount S. (1994)   ():`. That shape comes from the API and is tracked separately.

Please review on the preview or stage, **not test** — the test API deploy predates agr_java_software#1662 and silently ignores `filter.referenceCitation`, so the filter box there returns every row.

cerivs:
Looks good on the test site

cerivs:
Looks good, trying to get the code review to work

cerivs (approved):
Looks good on the test site.

PR #1863, opened 2026-08-19
fix(interactions): render every reference in the Reference ID column (KANBAN-1486)

jimflybase:
Routes both gene-page interactions tables' **Reference ID** column through `ReferencesCellCuration`, so an interaction supported by several publications shows all of them. Jira: [KANBAN-1486](https://agr-jira.atlassian.net/browse/KANBAN-1486)

## Summary

Both tables built that cell by hand from `evidence[0]` and discarded the rest of the array. The other seven tables with a Reference ID column (phenotypes, expression, disease annotations, DiseaseToGene / DiseaseToAllele / DiseaseToModel, AlleleToDisease) all render it with `ReferencesCellCuration`, which de-duplicates and wraps the list in a `CollapsibleList` ("Show All N", first two shown). Interactions was the only holdout, and the only one with a bespoke formatter.

A small helper, `evidenceReferenceIds`, adapts the shape: the interactions endpoints carry publications in `evidence[]` as `referenceID` plus a nested `crossReferences` list, whereas the other tables receive a ready-made array of cross references. It pulls out the cross reference whose `referencedCurie` matches `referenceID` — the same one the old code linked — and falls back to `{ referencedCurie }` when none matches, which renders the id unlinked exactly as before.

Removing the hand-rolled formatters also removes their stale `// eslint-disable-next-line react/prop-types` directives, so `src/components/interaction/` now lints clean (those were the 2 pre-existing errors noted in #1862).

## No visible change on today's data

The ticket asked for a targeted query before sizing this. I sampled **3,174 evidence entries** across 8 genes (human TP53, WB `daf-2`, fly `w`, mouse, yeast, zebrafish, rat) on both endpoints on stage: **every interaction row carries exactly one evidence entry**, so nothing on a page looks different today. This is alignment and future-proofing — worth knowing if you'd rather it wait.

## Verification

- New unit tests (9) cover the multi-evidence case that no live row provides: all references rendered and linked, `Show All 3` past two, single-evidence unchanged, unlinked fallback, and empty input.
- Browser check on `/gene/FB:FBgn0003996` against the stage API: both tables' Reference ID cells render the same single link as before, PMID hrefs intact (`PMID:26527002`, `PMID:21912621`, …), and the Reference column is untouched.
- `tsc --noEmit` clean; Jest 45 suites / 204 tests pass; Prettier clean; ESLint clean on the changed directory.

[KANBAN-1486]: https://agr-jira.atlassian.net/browse/KANBAN-1486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
+++ b/src/components/interaction/GeneGeneticInteractionDetailTable.jsx
@@ -1,11 +1,11 @@
 import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
-import { DataTable, GeneCellCuration, ReferenceList, SpeciesCell } from '../dataTable';
+import { DataTable, GeneCellCuration, ReferenceList, ReferencesCellCuration, SpeciesCell } from '../dataTable';
 import { getIdentifier } from '../dataTable/utils.jsx';
-import { buildUrlFromTemplate } from '../../lib/utils.js';
 import SpeciesName from '../SpeciesName.jsx';
 import ExternalLink from '../ExternalLink.jsx';
 import DataSourceLinkCuration from '../dataSourceLinkCuration.jsx';
+import evidenceReferenceIds from './evidenceReferenceIds.js';
 import MITerm from './MITerm.jsx';
 import useDataTableQuery from '../../hooks/useDataTableQuery';
 import AlleleCellCuration from '../dataTable/AlleleCellCuration.jsx';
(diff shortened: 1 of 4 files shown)


jimflybase:
I'm going to self-approve this PR, as it is kind of an infrastructure thing right now (no current instances of the problem it fixes).  As mentioned above, this is future-proofing.

PR #1875, opened 2026-08-20
SCRUM-6438 use projected allele viewer identifiers

christabone:
## Summary

- replace the full allele-table viewer request with the projected `allele-viewer-ids` contract
- skip viewer-summary requests entirely for human and SGD ISOFORM pages
- key viewer queries only by gene and effective filters, not table page/size/sort
- fetch every bounded viewer-ID page sequentially without changing unrelated orthology, paralogy, or reference-count consumers
- derive selected-row allele categories from the separately fetched variant list
- pin alliance-genome/genomefeatures#18395834ec62270dedcb09bce0747f81b09bb39e for explicit empty-filter semantics

## Verification

- 14 final-scope focused Jest tests pass across viewer paging/filtering and selected-row behavior
- full `npm run build` succeeds
- Prettier, install/build, and security CI pass on `789d46d2`

The paired API work in alliance-genome/agr_java_software#1664 and the focused library change in alliance-genome/genomefeatures#6 are merged. This UI PR still requires joint stage validation with those changes.


christabone:
@claude Please re-review the latest head. The prior review led to stable viewer mounting during selection/page loading, an orthology paging-error guard and regression test, correct zero-allele ISOFORM behavior, and a shared ISOFORM-only taxon constant.

christabone:
@claude Please confirm the latest head. The dependency array now tracks the object actually read, the dead dependency is removed, and placeholder IDs are retained only when the prior query belongs to the same gene; a focused regression test covers same-gene versus cross-gene behavior.

christabone:
@claude Please review the latest head (`6ed4aee7`). It includes the memo/placeholder fixes requested in the prior re-review plus migration of the remaining checked-in `limit=0` reference-count request to `limit=1`, preserving its use of the response total under the approved API minimum.

christabone:
@claude Please re-review the final narrowed head `789d46d2`. Maintainer feedback on the paired API PR removed the global SCRUM-6437 pagination policy, so this UI head restores the existing orthology/paralogy `limit=10000` behavior, restores reference counts to intentional `limit=0`, and removes the orthology-only error guard/test. `fetchAllPages` now serves only the new bounded allele-viewer endpoint. Please verify that the retained SCRUM-6438 viewer changes remain correct and that the SCRUM-6437 cleanup left no stale coupling.

christabone:
@claude Addressed the duplicated category strings in `43657685` by reusing `ALLELE_WITH_ONE_VARIANT` and `ALLELE_WITH_MULTIPLE_VARIANTS`; the 14 focused tests and Prettier check pass. I also confirmed the paired merged API contract: `/allele-viewer-ids` accepts `limit=1000`, filters to `allele_summary` before search, maps exactly one identifier per ES hit without deduplication, and reports that same search total, so `fetchAllPages` page offsets and completeness validation are consistent. The extreme initial-load latency remains documented follow-up material rather than a blocker. Please confirm the small constants cleanup if needed.

christabone:
Final Claude-feedback disposition for head `43657685`:

- fixed viewer unmount/state loss during selection and selection-fetch failure
- fixed stale memo dependencies and cross-gene placeholder leakage
- preserved zero-allele human/SGD behavior and centralized the ISOFORM-only taxon set
- removed stale `useAllVariants` coupling; `fetchAllPages` now has one production consumer, the viewer-ID hook
- confirmed the merged API accepts `limit=1000`, returns one ID per matching hit without deduplication, and reports the same query total; Claude considers the paging-contract question closed
- replaced duplicated allele-category strings with the existing constants in `43657685`
- intentionally reverted orthology/paralogy/reference-count changes because they belonged to the withdrawn SCRUM-6437 global policy; Claude confirmed exact parity with `stage`
- documented serial initial loading and extreme deep paging as non-blocking scale follow-ups to verify on stage; measured viewer-visible cardinality is currently 103 at the largest sampled gene
- retained pre-existing orthology error handling, paralogy-without-id behavior, and legacy `alleleTable` lint findings outside this focused PR

Final verification: GitHub Prettier/install-build/security checks pass; Claude ran `npm ci`, production build, 32 focused tests, Prettier, and touched-file lint, and marked the head LGTM. Joint stage validation remains the only project gate after maintainer review/merge.

jimflybase:
I got some help here from Claude; it recommends that I approve, but has comments:

  **Smaller things worth mentioning, none blocking**
  
  - formatAllele = (alleleId) => ({ id: alleleId }) correctly matches the endpoint's string results — I checked, because
    a shape mismatch there would have silently emptied the viewer.
  - hasManyAlleles (>20,000) still only disables the download button, and given the projection sizes above the viewer
    path needs no equivalent guard. Fine, just worth knowing it's unguarded by design.
  - ISOFORM_ONLY_TAXA hardcodes NCBITaxon:9606 / NCBITaxon:559292. It's an improvement over the inline comparison it
    replaces, and VariantsSequenceViewer now imports it — but taxon constants live in src/constants.js, so that's
    arguably the better home.
  - getViewerAlleleIdsUrl builds its query by splitting buildTableQueryString on & and keeping filter.*. Works while
    values stay URL-encoded; slightly brittle if a filter ever emits a raw &.
  - placeholderData keeps the previous gene's IDs only when geneId matches, so filter changes show briefly stale viewer
    content rather than flashing empty — deliberate and the right call.

  **One question I'd ask Chris**, not an objection: rat RGD:2917 has 916 alleles but 0 viewer IDs, and zebrafish likewise.
  Those genes will render "No mapped variant information available". I believe that matches the old behavior — the
  previous code derived hasVariants from variantList across filtered alleles, which would also have been empty — but
  that's the one behavior I couldn't confirm didn't change, since the old code path is gone from the branch.

christabone:
@jimflybase Thank you — your Rat question uncovered a real viewer-mounting regression. Stage confirms `RGD:2917` has 0 mapped allele IDs but 916 standalone variant rows, so the old UI mounted the viewer and the prior PR head did not. I updated this PR in `82edbfa3` and opened the paired draft API follow-up alliance-genome/agr_java_software#1665. The API keeps the bounded allele-ID results and adds a filtered `hasStandaloneVariants` signal, so the UI preserves the viewer without trying to page through the standalone/HTP HGVS set (which can approach 99,000 records for a gene and is not the embedded allele-associated VCF). I also moved `ISOFORM_ONLY_TAXA` into `src/constants.js` and replaced the split/reassemble URL handling with `URLSearchParams`, with an encoded `&`/`=` regression test.

@claude Please re-review latest head `82edbfa3` against paired API PR alliance-genome/agr_java_software#1665, including the standalone-only, mapped-allele, filter, selection, empty-result, paging-metadata, taxon, and encoded-filter cases.

christabone:
@claude Thank you for the thorough pass. I agree the undefined-versus-empty choice must be deliberate, but the claim that stage draws all 916 `RGD:2917` records is not supported by the actual embedded data source. I queried the exact gene interval (`12:5981835-6129275`) through the same `genomefeatures.fetchTabixVcfData` function against both `VCF/8.2.0/rat-latest.vcf.gz` and `VCF/9.0.0/rat-latest.vcf.gz`; both return **0 VCF records**. The 916 rows are standalone/high-throughput `variant_summary` table/index records, not records in the embedded allele-associated VCF. Thus stage mounts the viewer because old `hasVariants` sees their `variantList`, but it has zero variant glyphs at this locus. Head `82edbfa3` restores that mounted isoform/viewer shell and also draws zero variant glyphs.

Passing `undefined` would not draw those 916 records; it would show every allele-associated VCF record in the region, including records outside the filtered ID contract. That also conflicts with the existing viewer help text: “Only variants associated to alleles are shown.” The explicit-empty library fix was introduced precisely to prevent an empty resolved filter from becoming “show all,” so I am retaining `[]` here.

One other factual correction: the URL cleanup cannot be reduced to the direct `buildTableQueryString` call under current code. The project uses `qs`, whose option is `skipNulls`; `buildTableQueryString` currently passes `skipNull`, and the focused test demonstrated the direct form emits `page=&limit=&sortBy=`. The `URLSearchParams` deletion loop therefore is not a no-op. Its `+` encoding was verified by you against stage and is safe.

I addressed the API review test gap in agr_java_software#1665 head `10e6285ee` by restoring `assertFalse(identifierQuery.contains("variant_summary"))`; 7 focused service tests pass. Please re-evaluate the standalone-only disposition with the actual Tabix result and confirm whether any code issue remains.

christabone:
@claude I checked the remaining low-confidence intersection directly against the stage `site_index` alias: `category=allele_summary` + `exists(variantList)` + `alterationType NOT IN (allele with one variant, allele with multiple variants)` returns exactly 0 documents across all 24 shards (query took 97 ms). So mapped allele documents carrying variants are exhausted by the visible-ID alteration types, and there is no third mounting gap. Thank you for rechecking the Tabix path and confirming LGTM.

christabone:
@jimflybase The edge you identified is now fully addressed at `82edbfa3`, and the paired API follow-up is alliance-genome/agr_java_software#1665 at `3a62e2218`. The UI again mounts the viewer for standalone-only records while retaining an explicit empty VCF visibility filter.

We verified the behavior directly: all 916 standalone `RGD:2917` table rows have mapped HGVS locations, but the exact gene interval (`12:5981835-6129275`) contains 0 records in both the embedded Rat 8.2.0 and 9.0.0 VCF files. Thus stage and this head both mount the viewer shell and draw zero variant glyphs there; passing an unfiltered VCF would risk showing allele-associated records from neighboring/overlapping genes instead. A stage-wide ES audit also found zero variant-bearing `allele_summary` documents outside the one/multiple-variant alteration types.

The UI checks are green (5 suites / 33 tests, production build, formatting, and GitHub CI), and Claude independently repeated the Tabix verification, retracted its earlier concern, and marked the head LGTM. When you have a chance, could you please give the latest UI head a final review?

christabone:
@jimflybase Quick review update: I have merged the latest `stage` into this branch without conflicts. The current PR head is `46d0f62b`; the viewer behavior remains the reviewed `82edbfa3` change. After the merge update, the 5 focused suites / 33 tests, touched-file formatting, and production build all pass locally. GitHub CI is now rerunning on the refreshed head.

PR #1880, opened 2026-08-20
feat(diseasePortal): Disease Portal page edits for release 9.1.0 (KANBAN-1493)

jimflybase:
All Disease Portal changes for release **9.1.0**, requested at the Aug 19 D&P meeting and tracked under [KANBAN-1493](https://agr-jira.atlassian.net/browse/KANBAN-1493). Nine tickets, twelve commits, nine files.

**Curator-approved.** Ranjana reviewed the whole set together on the shared Amplify preview (`feature-kanban-1493-dp-9-1-0`) on Aug 20 2026 and approved all changes, including the Recent Literature hide. She has no GitHub account, so this is the record of that sign-off.

## What's in it

| Ticket | Change |
| --- | --- |
| [1492](https://agr-jira.atlassian.net/browse/KANBAN-1492) | Diseases headline pill hidden on all portal pages |
| [1495](https://agr-jira.atlassian.net/browse/KANBAN-1495) | Ontology View moved directly below Summary |
| [1494](https://agr-jira.atlassian.net/browse/KANBAN-1494) | Recent Papers moved to the bottom, renamed Recent Literature |
| [1496](https://agr-jira.atlassian.net/browse/KANBAN-1496) | Browse Ontology removed from the portal side nav |
| [1497](https://agr-jira.atlassian.net/browse/KANBAN-1497) | "Browse ontology for &lt;disease&gt;" link on the Ontology View heading; tree term labels unlinked; link toned down after preview feedback |
| [1498](https://agr-jira.atlassian.net/browse/KANBAN-1498) | Disease Portal link at the bottom of the Summary on disease pages |
| [1499](https://agr-jira.atlassian.net/browse/KANBAN-1499) | Portal headline number links open in a new tab |
| [1488](https://agr-jira.atlassian.net/browse/KANBAN-1488) | Index category headings sorted alphabetically |
| [1489](https://agr-jira.atlassian.net/browse/KANBAN-1489) | Community Resources: full name first, acronym in parentheses; SADS Foundation → The Foundation for Inherited Arrhythmias (FIA), URL included |
| [1473](https://agr-jira.atlassian.net/browse/KANBAN-1473) | Recent Literature section hidden for this release |

Net effect on a portal page: **Summary → Ontology View → Community Resources**, no Diseases pill, no Browse Ontology nav item, counts opening in new tabs.

## Things worth knowing before merging

[...]

PR #1881, opened 2026-08-20
docs(ontologyBrowser): correct the nodeHref comment, now that no caller passes it

jimflybase:
Comment-only fix in a shared component, from the bot review on #1880. Targets `stage` directly — `OntologyTree.jsx` was not among that PR's files, so there is no overlap.

The comment said embedded viewers link every label out and "the standalone omits `nodeHref`". After [KANBAN-1497](https://agr-jira.atlassian.net/browse/KANBAN-1497) the disease-portal embed omits it too, so **no caller passes it at all** — the comment described behavior that had just been removed, in the one component shared by both consumers. Left alone it reads as an invitation to restore the links, which is exactly the trap the epic fixed in `RECENT_PAPERS_API.md`.

Now documents the prop as an opt-in seam, and why it is kept rather than deleted: the decision has already reversed twice — KANBAN-1470 restored these links, KANBAN-1497 removed them again — so the seam is worth keeping for the same reason the pill and literature hides sit behind switches. Also points at [KANBAN-1500](https://agr-jira.atlassian.net/browse/KANBAN-1500) for keyboard selection, since the label link used to be the row's only focusable element.

No behavior change. Prettier and ESLint clean.

[KANBAN-1497]: https://agr-jira.atlassian.net/browse/KANBAN-1497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KANBAN-1500]: https://agr-jira.atlassian.net/browse/KANBAN-1500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/ontologyBrowser/OntologyTree.jsx
+++ b/src/containers/ontologyBrowser/OntologyTree.jsx
@@ -70,9 +70,14 @@ const OntologyTree = ({
   // loses it a moment later when the batched fetch reports zero children.
   const hasChildren = childTerms.length > 0 || (data?.doTerm?.descendantCount || 0) > 0;
   const isFocused = focusedCurie === curie;
-  // Embedded viewers (the disease portal) link every label out to the full
-  // browser, since they have no detail panel of their own. The standalone
-  // omits nodeHref and keeps rendering a plain span that selects in place.
+  // nodeHref is an opt-in seam: supply it and the term label renders as a link
+  // out to the full browser instead of a plain span that selects in place. No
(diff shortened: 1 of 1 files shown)


jimflybase:
Again self-approving, as this is all comments that most likely only I will ever read anyway.

PR #1885, opened 2026-08-21
fix(dataTable): don't link uncited or non-AGRKB references (KANBAN-1505)

jimflybase:
Fixes [KANBAN-1505](https://agr-jira.atlassian.net/browse/KANBAN-1505).

`ReferenceList` backs the "Reference" column in all 8 KANBAN-864 tables. It rendered `{ref.shortCitation}` inside `<Link to={/reference/${ref.curie}}>` with no guard, so annotation evidence that is an `ExternalDatabaseReference` (Orphanet, OMIM) — which has neither a citation nor an Alliance reference page — produced an **empty link to a URL that 400s**.

Most visible on human gene pages. On `/api/gene/HGNC:11998/phenotypes` (TP53), **73 of 81 references** are uncited ORPHA entries; only 8 are real literature. The column read as blank enough to look missing.

### The fix

Ports the two guards `singleReferenceLinkCuration.jsx` already applies (which is why the three annotation popups were never affected):

1. no `shortCitation` → the entry is skipped
2. curie not `AGRKB:` → the citation renders as plain text, not a link

Uncited entries are filtered *before* `CollapsibleList` rather than returning `null` inside the map, so they no longer inflate its "Show All N" count.

**No API or Elasticsearch change is possible here.** For those 73 the evidence genuinely is the Orphanet entry — the annotation's identity is keyed on it (`uniqueId: HGNC:11998|has_phenotype|Abdominal obesity|HP:0012743|ORPHA:96253`) — so there is no Alliance reference to point at.

### Verified

| Page | Before | After |
| --- | --- | --- |
| TP53 (human) Phenotypes | 5/5 sampled rows had an empty `/reference/ORPHA:…` link | 0 empty links; cells cleanly blank, Reference ID still links to orpha.net |
| Trp53 (mouse) Phenotypes | citations linked | unchanged — "Hamard PJ (2013) Genes Dev 27(17):1868-85" → `/reference/AGRKB:101000000264700` |

8 new unit tests covering both guards, the mixed cited/uncited case, the Show All count, and filter dimming. Full suite 49 suites / 227 tests green; ESLint and Prettier clean.

[KANBAN-1505]: https://agr-jira.atlassian.net/browse/KANBAN-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/dataTable/referenceList.jsx
+++ b/src/components/dataTable/referenceList.jsx
@@ -14,26 +14,31 @@ const ReferenceList = ({ refs, trunc = true, filterTerm }) => {
         .toLowerCase()
     )
     .filter(Boolean);
+  // Orphanet/OMIM evidence carries no citation, so there is nothing to show (KANBAN-1505).
+  const citedRefs = (refs || []).filter((ref) => ref?.shortCitation);
+  if (!citedRefs.length) {
+    return null;
+  }
+
   return (
-    refs && (
-      <CollapsibleList>
-        {refs.map((ref) => {
-          // when a filter is active, de-emphasize references matching none of its terms
-          const citation = (ref.shortCitation || '').toLowerCase();
-          const dim = terms.length > 0 && !terms.some((term) => citation.includes(term));
-          return (
-            <Link
-              to={`/reference/${ref.curie}`}
-              title={ref.shortCitation}
-              className={`${trunc ? style.ellipses : ''}${dim ? ' ' + style.referenceDim : ''}`}
-              key={ref.curie}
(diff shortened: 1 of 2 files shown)


jimflybase:
Another mostly invisible fix.  The PR claim that it's visible in Hsap TP53 Phenotypes tables . . . well, it's not easy to see even there.  Self-approving.

PR #1888, opened 2026-08-21
fix(sitemap): add the application routes and derive portal URLs from portalData (KANBAN-1472)

jimflybase:
Fixes [KANBAN-1472](https://agr-jira.atlassian.net/browse/KANBAN-1472).

`bin/build_sitemap.js` was a hand-maintained list of 17 URLs, every one a WordPress content page. No application route had ever been added, so four shipped features were invisible to search engines — including the eight curator-authored disease portal pages, whose entire purpose is to be found.

**17 → 30 URLs.** Diffed the generator output against `stage`; 13 added, none removed:

```
/disease-portal                            /downloads
/disease-portal/alzheimers-disease         /news
/disease-portal/autism-spectrum-disorder   /ontology/disease
/disease-portal/ciliopathy                 /blastservice
/disease-portal/colorectal-cancer
/disease-portal/diabetes-mellitus
/disease-portal/epilepsy
/disease-portal/long-qt-syndrome
/disease-portal/parkinsons-disease
```

### Portal URLs are derived, not listed

They come from `portalData` via `Object.entries(...).filter(([, p]) => !isRootPortal(p))`, so the next portal is indexed the moment it is added. Hardcoding is precisely how the existing eight drifted out of the sitemap. `isRootPortal` drops the root entry, which is the `/disease-portal` index page itself and contributes no slug.

### Judgement calls

- **`/search` stays out** — search result pages are not worth indexing.
- **`/ontology` is a client-side `<Navigate replace to="/ontology/disease">`** (`routes.jsx:55`), so the target is listed rather than the redirect.
- Also drops the unused `fs` import, which ESLint was flagging on this file.

### Verified

- `npm run sitemap` runs clean; output is well-formed per `xmllint`. The import chain `build_sitemap.js → portalData.js → ontologies.js` is plain JS with no further imports, so it is safe under plain `node`.
- All four new non-portal routes render: `/ontology/disease` ("Ontology Browser"), `/blastservice` ("BLAST Service"), `/news` ("News and Events"), `/downloads` ("Downloads").
- ESLint and Prettier clean; full suite green at 49 suites / 227 tests.

### Two things found while in here, both worth knowing and neither changed

1. **The serving chain is fine.** `public/sitemap.xml` is a tracked, hand-written index pointing at `/sitemap-index.xml` and `/api/sitemap.xml`; the first two generated files are gitignored and built at `prebuild`. So these URLs do reach the served sitemap, despite `sitemap.xml` itself carrying a Nov 2025 timestamp.
2. **Entity pages are already covered, by the API.** `/api/sitemap.xml` is a 343-entry sitemap index — 245 allele, 97 gene, 1 disease. That answers the ticket's open question about whether genes/diseases/alleles belong here: they don't, they are already indexed server-side. It also scopes KANBAN-1480 usefully — reference and variant sitemaps are absent from that API index, so that work belongs in agr_java_software, not here.

---

## Second commit: member landing pages (30 → 38)

Adds the 8 MOD landing pages, derived from `NAV_MENU` rather than a second hardcoded list. This settles the `/members/:id` question that [KANBAN-1480](https://agr-jira.atlassian.net/browse/KANBAN-1480) AC 3 left to "here or in KANBAN-1472" — it needs no API work, unlike the reference and variant sitemaps that ticket also covers.

```
/members/flybase   /members/sgd        /members/zfin
/members/mgd       /members/wormbase   /members/goc
/members/rgd       /members/xenbase
```

### This required moving RECAPTCHA_SITE_KEY

`constants.js` could not be imported from a plain `node` script at all. Line 3 evaluated `import.meta.env.VITE_RECAPTCHA_SITE_KEY` at module scope, and outside Vite `import.meta.env` is `undefined`, so the import threw:

```
TypeError: Cannot read properties of undefined (reading 'VITE_RECAPTCHA_SITE_KEY')
    at src/constants.js:3:51
```

`RECAPTCHA_SITE_KEY` moves to `contactPage/index.jsx`, its only consumer, leaving `constants.js` as pure data any generator can import. The alternative — hardcoding eight URLs — would have contradicted the whole point of this PR.

**The expression itself is unchanged, only relocated**, so Vite's define-substitution behaves exactly as before. Verified rather than assumed: a full production build with `VITE_RECAPTCHA_SITE_KEY=test-key-sentinel-12345` puts that sentinel in the bundle **once**, with **no** unsubstituted `VITE_RECAPTCHA_SITE_KEY` remaining. `npm run build` succeeds end to end, prebuild generators included.

[KANBAN-1472]: https://agr-jira.atlassian.net/browse/KANBAN-1472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[KANBAN-1480]: https://agr-jira.atlassian.net/browse/KANBAN-1480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/build_sitemap.js
+++ b/bin/build_sitemap.js
@@ -1,5 +1,6 @@
 import { simpleSitemapAndIndex } from 'sitemap';
 import { data as portalData, isRootPortal } from '../src/containers/diseasePortal/portalData.js';
+import { NAV_MENU } from '../src/constants.js';
 
 // derived from portalData so a new portal is indexed without anyone remembering this file
 const diseasePortalUrls = Object.entries(portalData)
@@ -10,6 +11,14 @@ const diseasePortalUrls = Object.entries(portalData)
     priority: 0.7,
     lastmodrealtime: true,
   }));
+
+// likewise from the nav, so a new MOD landing page is indexed with no second edit
+const memberUrls = (NAV_MENU.find((section) => section.label === 'Members')?.sub ?? []).map(({ route }) => ({
+  url: route,
+  changefreq: 'monthly',
+  priority: 0.7,
+  lastmodrealtime: true,
+}));
 
 simpleSitemapAndIndex({
   hostname: 'https://www.alliancegenome.org',
@@ -40,5 +49,6 @@ simpleSitemapAndIndex({
     { url: '/downloads', changefreq: 'monthly', priority: 0.7, lastmodrealtime: true },
     { url: '/news', changefreq: 'weekly', priority: 0.7, lastmodrealtime: true },
     { url: '/blastservice', changefreq: 'monthly', priority: 0.7, lastmodrealtime: true },
+    ...memberUrls,
   ],
 });
--- a/bin/build_sitemap.js
+++ b/bin/build_sitemap.js
@@ -1,5 +1,15 @@
 import { simpleSitemapAndIndex } from 'sitemap';
-import fs from 'fs';
+import { data as portalData, isRootPortal } from '../src/containers/diseasePortal/portalData.js';
+
+// derived from portalData so a new portal is indexed without anyone remembering this file
+const diseasePortalUrls = Object.entries(portalData)
+  .filter(([, portal]) => !isRootPortal(portal))
+  .map(([slug]) => ({
+    url: `/disease-portal/${slug}`,
+    changefreq: 'monthly',
+    priority: 0.7,
+    lastmodrealtime: true,
+  }));
 
 simpleSitemapAndIndex({
   hostname: 'https://www.alliancegenome.org',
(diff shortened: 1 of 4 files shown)


jimflybase:
@christabone added here in case you don't agree with the scope of pages covered by the sitemap doc - I wonder in particular if the previous omission of the Downloads page from that doc might have been intentional.  @Joe-Wdz added because you may take up the remaining related tickets; you should know what's been done.

jimflybase:
Both points are now addressed — the review landed on the first commit, 23 minutes before the second one was pushed.

**1. `/members/*` — done, and done the way you suggested** (commit `eefbae59`). Derived from `NAV_MENU` rather than hardcoded, so all eight MOD landing pages are emitted and a ninth would be picked up automatically. Sitemap is now 38 URLs.

That turned out to need more than an import. `constants.js` **could not be imported under plain `node` at all** — line 3 evaluated `import.meta.env.VITE_RECAPTCHA_SITE_KEY` at module scope, and outside Vite `import.meta.env` is `undefined`:

```
TypeError: Cannot read properties of undefined (reading 'VITE_RECAPTCHA_SITE_KEY')
    at src/constants.js:3:51
```

So `RECAPTCHA_SITE_KEY` moved to `contactPage/index.jsx`, its only consumer. The expression text is unchanged, only relocated, and I verified the substitution rather than assuming it: a production build with `VITE_RECAPTCHA_SITE_KEY=test-key-sentinel-12345` puts that sentinel in the bundle exactly once, with no unsubstituted `VITE_RECAPTCHA_SITE_KEY` left.

**2. The `bin/` → `src/` build-time coupling — noted in both files** (commit `5047a2d8`). Your point is sharper than hypothetical: `constants.js` had exactly that defect until this branch, and it would have taken the whole Amplify deploy down rather than just sitemap generation, since `prebuild` gates `vite build`. Both `portalData.js` and `constants.js` now carry a header saying they are imported by the build script and must stay free of JSX, TypeScript and `import.meta`.

Also confirming your `/search` observation independently: leaving it out matches `Disallow: /search` in `bin/build_robots.js`, so the two generators now agree rather than contradicting each other.

christabone (approved):
Reviewed the current three-commit head. KANBAN-1472 explicitly includes `/downloads`, so that route belongs in scope. The generated portal and member URLs match live routes, the recaptcha-key move preserves its only consumer while making the generator import chain Node-safe, and the full install/test/build plus formatting and secret checks are green. No blocking issues found.

PR #1884, opened 2026-08-21
feat(diseasePortal): order pills Genes, Alleles, Models and fix the Species tooltip (KANBAN-1503)

jimflybase:
Curator request from the [KANBAN-1471](https://agr-jira.atlassian.net/browse/KANBAN-1471) thread, tracked as KANBAN-1503 under the Disease Portal 9.1.0 epic.

**Pill order.** Sue Bello: *"On the disease pages and in the browser, we go Gene, Allele, Model. I think it would be helpful to keep this order consistent."* The portal rendered Models, Genes, Alleles in both places she named:

- `DiseasePortalSection.jsx` — the coloured top-section pills
- `PortalListSection.jsx` — the count line beside each portal link on `/disease-portal`

**Species tooltip.** The pill's hover text read "View all associated species" but the link goes to `/about-us`. Sue flagged it as misleading and Ranjana approved the change to "About the Alliance".

The hidden Diseases pill (KANBAN-1492) is unaffected, and Species stays last on the root portal.

### Verified locally

| Page | Result |
| --- | --- |
| `/disease-portal` | Genes, Alleles, Models, Species |
| `/disease-portal/epilepsy` | Genes, Alleles, Models |
| Landing count lines | "4,167 genes / 34 alleles / 86 models" |
| Species tooltip | "About the Alliance", href still `/about-us` |

Hrefs and `target="_blank"` (KANBAN-1499) are unchanged. Jest 48 suites / 219 tests pass; ESLint and Prettier clean.

No test updates needed — the Playwright coverage in agr_release_testing asserts per-link and per-label, not left-to-right order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[KANBAN-1471]: https://agr-jira.atlassian.net/browse/KANBAN-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1889, opened 2026-08-21
fix(diseasePortal): render a zero count as 0 instead of a blank pill (KANBAN-1504)

jimflybase:
Fixes [KANBAN-1504](https://agr-jira.atlassian.net/browse/KANBAN-1504), raised by the automated review on #1884 and deliberately left out of that PR.

The headline numbers in `DiseasePortalSection.jsx` guarded on truthiness:

```jsx
<div>{alleleCount ? alleleCount.toLocaleString() : ''}</div>
```

A genuine count of **0** is falsy, so the pill rendered its label with an empty number above it — visually identical to a count that failed to load. A user could not tell "this disease has zero alleles" from "this number is broken". `PortalListSection.jsx` already guarded the same data with `!= null` and would have said "0 alleles", so the two components disagreed.

`useEntityButtonCounts` returns `undefined` while the query is in flight, which `!= null` already covers — so the truthiness check bought nothing and only ever cost the zero case.

### All four pills, including the hidden one

Genes, Alleles and Models, plus the **Diseases** pill that KANBAN-1492 currently hides behind `SHOW_DISEASE_COUNT`. It carried the same guard and would have inherited the bug the day that flag flips back on — fixing three of four and leaving a latent one behind a feature flag is how this sort of thing survives.

### Tests

First tests for this component: zero renders `0`, `undefined` renders nothing (and no stray `false`), and a populated count still thousands-separates.

**Confirmed they actually catch the regression.** Reverted one guard to the old form and re-ran:

```
Expected: "0Genes"
Received: "Genes"
```

That matters here because no portal is anywhere near zero — the floor across all eight is 17 alleles on Long QT — so this path is unreachable from live data and a passing test proves nothing on its own.

Full suite green at 50 suites / 230 tests; ESLint and Prettier clean. No visible change on any current page.

[KANBAN-1504]: https://agr-jira.atlassian.net/browse/KANBAN-1504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/containers/diseasePortal/DiseasePortalSection.jsx
+++ b/src/containers/diseasePortal/DiseasePortalSection.jsx
@@ -22,7 +22,7 @@ const DiseasePortalSection = ({ disease }) => {
       `${disease.pageName} Portal`;
   const diseasesEntityButton = SHOW_DISEASE_COUNT ? (
     <EntityButton id="entity-diseases" to="/search?q=&category=disease_search_result" tooltip="View all diseases">
-      <div>{diseaseCount ? diseaseCount.toLocaleString() : ''}</div>
+      <div>{diseaseCount != null && diseaseCount.toLocaleString()}</div>
       Diseases
     </EntityButton>
   ) : (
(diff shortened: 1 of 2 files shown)


oblodgett (approved):

jimflybase:
This one is insurance - we are unlikely to ever expose a Disease Portal page with zero curated data instances for one of the G/A/M data types, but if we ever do this would properly render the corresponding pill.  No review possible, unless someone wants to mess with devtools.  Self-approving.

PR #1894, opened 2026-08-31
SCRUM-6460 Fix allele viewer table filtering

christabone:
## Summary

- preserve the nested allele API response used by the gene table so viewer-selected allele IDs continue to match table row IDs
- unwrap variant objects from the allele variant-summary response so selected rows retain HGVS, variant type, consequence, and JBrowse location data
- add focused regression coverage for both API-to-table shape contracts

## Validation

- `npm test -- --runInBand` (50 suites, 232 tests passed)
- `npm run build`
- targeted ESLint and Prettier checks for both changed files
- Chrome smoke test against stage data for `MGI:105043`: selecting the `Ahrb-2` glyph produced one table row with `NC_000078.7:g.35550690A>G`, `point mutation`, `stop lost`, and a valid JBrowse highlight
- independent GPT-5.6 Sol high review: accepted with no findings after the final browser-tested revision

Jira: https://agr-jira.atlassian.net/browse/SCRUM-6460

### DIFF
--- a/src/hooks/tests/useAlleleSelection.test.js
+++ b/src/hooks/tests/useAlleleSelection.test.js
@@ -1,4 +1,5 @@
-import { getSelectedAlleleCategory } from '../useAlleleSelection';
+import { buildSelectedAlleleRow, getSelectedAlleleCategory, getSelectedVariantList } from '../useAlleleSelection';
+import { getIdentifier } from '../../components/dataTable/utils';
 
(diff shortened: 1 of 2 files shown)


christabone:
@claude Please review my changes. I applied your consistency suggestion so selected-allele deduplication now uses the same getIdentifier accessor as the table row key. All 232 tests and targeted lint/format checks pass.

PR #1901, opened 2026-09-08
SCRUM-6529 Keep positive and NOT GO ribbon annotations separate

### LINKED ISSUE #1869: A possible issue in the grouping choice of the GO Ribbon, leading to unintuitive rendering
To duplicate, first we can see what the expected fields are in the API:

`curl -s 'https://api.geneontology.org/api/bioentityset/slimmer/function?slim=GO%3A0003824&subject=HGNC%3A5320&rows=-1' | jq -r '.[0].assocs[] | select(.object.id==["GO:0004415"](go:0004415)) | [(.negated|tostring), .evidence_type, (.reference|join(",")), (.provided_by|join(","))] | @tsv'`

This is 18 with no qualifier and 4 with NOT.

Now, when I go to the Alliance page https://www.alliancegenome.org/gene/HGNC:5320#function---go-annotations and click the "catalytic activity cell", I see that everything is listed as a NOT.

In all cases, it seems like the data is correct.

After a little probing around, we started thinking that this may be a frontend issue. Specifically, the usage of the `go-annotation-ribbon-table` component. I'm not sure of the history of why it is setup this way, but the UI is specifying a custom `group-by` attribute as just `term`: https://github.com/alliance-genome/agr_ui/blob/04469588fcecdb23dad98fdfcf34a197ab40f69c/src/components/geneOntologyRibbon/index.jsx#L551-L559

We think that it may be easiest not having an override here (as the default for `group-by` is `term,qualifier`, allowing qualifier to be exposed).

That said, we're also uncertain of the use case here.

Tagging @christabone

christabone:
The GO ribbon table merges positive and negated annotations for the same term, retaining the first annotation's term tags. For HYAL1 (HGNC:5320), clicking catalytic activity displays all 22 annotations for GO:0004415 under “not”, although the API returns 18 positive and 4 negated annotations.

Group by `term,qualifier` so each group retains its own term tags, evidence and references. The installed `@geneontology/web-components` 1.0.1 has no default `groupBy`, so the grouping is explicit. The qualifier continues to appear beside the term using the component's existing tags.

Tracking: [SCRUM-6529](https://agr-jira.atlassian.net/browse/SCRUM-6529). Fixes #1869.

Validation:

- Four regression tests exercise the installed GO component with the attributes supplied by the Alliance component: both input orders, evidence/reference separation, experimental filtering across qualifiers, and the all-annotations view. All four fail before the fix and pass afterward.
- Full Jest suite: 51 suites, 236 tests passed. Production build passed.
- Browser check using the live GO API response on 2026-09-08: the old grouping renders one NOT group containing all 22 annotations; the new grouping renders 4 NOT annotations and 18 positive annotations in separate groups.
- Changed files pass Prettier. ESLint reports the same 23 pre-existing errors in the component on the base branch and this branch; the new test has no lint errors.

Suggested fix for review by @kltm (Seth Carbon), @jimflybase (Jim Thurmond), and @alliance-genome/wg-ui.

[SCRUM-6529]: https://agr-jira.atlassian.net/browse/SCRUM-6529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/geneOntologyRibbon/index.jsx
+++ b/src/components/geneOntologyRibbon/index.jsx
@@ -552,7 +552,8 @@ class GeneOntologyRibbon extends Component {
         ref={this.tableRef}
         // bio-link-data={JSON.stringify(this.state.selected.data)}
         filter-by={this.state.onlyEXP ? 'evidence:' + EXP_CODES.join(',') : ''}
-        group-by="term"
+        // Keep positive and NOT annotations separate; the grouped term retains the first annotation's tags.
+        group-by="term,qualifier"
         // hide-columns={'qualifier,' + (this.state.selectedOrthologs.length == 0 ? 'gene,' : '') + (this.state.selected.group.id != 'all' ? ',aspect' : '')}
         hide-columns={'qualifier,gene,' + (this.state.selected.group.id !== 'all' ? ',aspect' : '')}
         order-by="term"
--- a/src/components/geneOntologyRibbon/qualifierGrouping.test.jsx
+++ b/src/components/geneOntologyRibbon/qualifierGrouping.test.jsx
@@ -1,0 +1,82 @@
+import GeneOntologyRibbonWithNavigate from './index.jsx';
+import { AnnotationRibbonTable } from '../../../node_modules/@geneontology/web-components/dist/collection/components/annotation-ribbon-table/annotation-ribbon-table.js';
+
+jest.mock('react-router-dom', () => ({ useNavigate: () => jest.fn() }));
+jest.mock('../OrthologPicker.jsx', () => () => null);
+// Exercise the installed component's real data conversion, grouping and filtering without network or Stencil rendering.
+jest.mock('@stencil/core', () => ({ h: jest.fn() }), { virtual: true });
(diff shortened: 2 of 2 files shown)


christabone:
Thanks for verifying the grouping behavior and reproducing the regression tests. No code changes are needed for the findings in this review.

The live reproduction has absent qualifiers on positive annotations and `["not"]` on negative annotations. The first-qualifier limitation belongs to the upstream grouping implementation; changing that behavior would be separate work if the endpoint supplies additional qualifiers.

The test deliberately exercises the installed component's real grouping implementation. `package-lock.json` currently resolves version 1.0.1, and the tests passed after `npm ci`. The internal import paths are a maintenance tradeoff: a future package layout change will require updating this test. No dependency pin or test restructuring is needed for this fix.

Local verification and GitHub build/test, formatting, and security checks all passed. Keeping the patch as reviewed.

This response was drafted by Codex.

oblodgett (approved):

